PR #850, opened 2026-05-18
feat(demo): add Chart Playground showcase with interactive knobs

alfredo1996:
## Summary

Adds a fifth seeded demo dashboard — **Chart Playground** — alongside the existing four showcases (Chart Gallery, Click Actions, Transformations, Rule-Based Styling). They co-exist; Chart Gallery is unchanged. The Playground is an interactive sandbox: every page surfaces one or more chart types together with **knobs** (parameter-select widgets) the viewer can twiddle to see how each chart responds.

## 8-page layout

1. **Overview** — markdown intro, three single-value KPIs (orders / products / customers).
2. **Categorical** — bar chart (dimension + metric + limit knobs, rule-based red styling for value > 5000) and pie chart, each with vertical/horizontal and solid/donut variant tiles.
3. **Time series** — line chart (time-grain + metric + window-days knobs) with straight-vs-smooth-area variants, plus gantt (limit knob) with progress-overlay variants.
4. **Distribution & Hierarchy** — treemap + sunburst + circle-packing in a single row, all consuming the same parameterized query so the three renderings are directly comparable.
5. **Geographic** — Leaflet map (continent filter + max-markers knob) with cluster/no-cluster variants, and a choropleth with metric switch.
6. **Network & Flow** — Neo4j graph (limit knob) and sankey (target-dimension switch).
7. **Single-value, Tabular & Specialty** — single-value row (metric switch, three format variant tiles), gauge (target-threshold knob, two display variants), table pair (server-side knobs + client-side transform comparison), JSON viewer (row-count knob), and a radar (region picker).
8. **Text & Interactive** — parameter-select wired to a single-value KPI **and** a markdown widget that interpolates the chosen value via `$param_*` substitution; form widget that writes to the demo feedback table with a read-back table beside it; iframe widget whose URL is parameter-driven.

## Parameter-substitution constraint

`$param_*` substitution only flows into widget queries, markdown `content`, and iframe `url` (see `app/src/components/card-container.tsx:421-434`). For chart options that can't be parameterized (palette, donut, smooth, etc.), the page renders **two side-by-side variant tiles** with one setting different so users can visually compare.

Numeric knobs use the `_max` companion of `number-range` parameters (e.g. `$param_cat_limit_max`) so the substitution gets a scalar `8` rather than the tuple `8,12` that the raw range param holds.

## Test plan

- [x] `node cli/dist/index.js demo seed --only chart-playground` succeeds
[...]

PR #851, opened 2026-05-18
fix(editor): expose number-range parameter type in widget editor

alfredo1996:
## Summary

The `number-range` parameter widget type is fully supported by the runtime (`ParamNumberRange` dual-handle slider, parameter-store, plugin Zod schema) but the widget editor never exposed it. Round-trip is broken:

- Users can't create a number-range parameter widget from the UI (no entry in the parameter-type dropdown).
- Existing number-range widgets reopen as type "Select" with no min/max/step controls — `reverseParamTypeMapping` falls through to the default case. Save-after-edit silently rewrites them to `select`.

## Root cause

`app/src/components/widget-editor/parameter-config-section.tsx`:
- `ParamUIType` only included `"date" | "freetext" | "select"`.
- `resolveInternalParamType` had no `"number-range"` branch.
- `reverseParamTypeMapping` had no `"number-range"` case.
- The parameter-type dropdown only listed three options.
- No min/max/step UI block.

The duplicated `reverseParamTypeMapping` in `app/src/stores/widget-editor-store.ts` (kept separate to avoid importing the UI barrel) had the same gap.

## Changes

- Add `"number-range"` to `ParamUIType` and both mappings in `parameter-config-section.tsx` + `widget-editor-store.ts`.
- Add "Number Range" (SlidersHorizontal icon) to the parameter-type dropdown.
- Add an editor UI block with min/max/step number inputs writing to `chartOptions.rangeMin/rangeMax/rangeStep` — same keys the runtime renderer (`parameter-widget-renderer.tsx`) and Zod schema (`plugins/parameter-select/settings.ts`) already consume.
- Add a `NumberRangeSlider` preview branch to `parameter-preview.tsx`.
- Update existing unit tests + add new cases: mapping round-trip, store `loadFromWidget` for number-range, `useBuildWidgetForSave` resolving `number-range` type with rangeMin/Max/Step preserved, and editor-UI smoke tests asserting min/max/step inputs render.

## Test plan

- [x] `npm -w app run test -- --run parameter-config-section use-widget-save widget-editor-store` — 136 tests pass.
- [x] Full `npm -w app run test` — all 2680 tests pass (one pre-existing transform error in `dashboard-error-boundary.test.tsx` from a missing `d3-hierarchy` dep, unrelated and present on `release/1.0`).
- [ ] Open a number-range widget from the Chart Playground (PR #850): type shows "Number Range", min/max/step inputs populated, slider preview renders.
- [ ] Create a new number-range parameter widget end-to-end: name, min=0, max=100, step=5; save; the rendered widget sets `$param_<name>`, `$param_<name>_min`, `$param_<name>_max`.

## Notes

Fixes the round-trip bug found while reviewing PR #850 (Chart Playground). Once this merges, the number-range widgets in Chart Playground will be editable from the UI.

The runtime accepts only `rangeMin`/`rangeMax`/`rangeStep` — no `defaultValue` prop exists on `ParamNumberRange`, so the editor doesn't offer one. The Chart Playground demo controls its initial value by pre-populating the parameter store, not via chartOptions; if a `defaultValue` knob is desired later it requires a separate runtime change.

The `form-fields-editor.tsx` number-range block uses a different component model (per-field flat object with `field.rangeMin` etc.) and is not directly reusable, so the parameter-config block is a small duplication of three labeled number inputs — not worth a shared component extraction for this size.

### DIFF
--- a/app/src/components/widget-editor/__tests__/parameter-config-section-ui.test.tsx
+++ b/app/src/components/widget-editor/__tests__/parameter-config-section-ui.test.tsx
@@ -90,7 +90,12 @@ vi.mock("@neoboard/components", () => ({
 
 vi.mock("lucide-react", () => {
   const Icon = () => <span />;
-  return { Calendar: Icon, Type: Icon, ListFilter: Icon };
+  return {
+    Calendar: Icon,
+    Type: Icon,
+    ListFilter: Icon,
+    SlidersHorizontal: Icon,
+  };
 });
 
 const mockSetParamUIType = vi.fn();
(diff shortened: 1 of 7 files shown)


alfredo1996:
Superseded by #853 — consolidated into a single PR for easier review (number-range editor + integer/float type + dual-handle slider fix + both demo dashboards).

Issue #917, opened 2026-06-03
[P1] Graph widget crashes on 'hierarchical' layout — schema enum + audit + graceful fallbacks


alfredo1996:
**Found during** [#895](https://github.com/alfredo1996/neoboard/issues/895) session 1a, opening an imported dashboard.

## Symptom (stack trace from runtime)
\`\`\`
[
  {
    "received": "hierarchical",
    "code": "invalid_enum_value",
    "options": ["force", "circular"],
    "path": ["layout"],
    "message": "Invalid enum value. Expected 'force' | 'circular', received 'hierarchical'"
  }
]
src/plugins/graph/component.tsx (36:40) @ GraphPluginComponent
  > 36 |   const settings = graphSettingsSchema.parse(raw);
\`\`\`

## Root cause — schema/feature mismatch
| File | Includes \`hierarchical\` |
|---|---|
| \`component/src/charts/graph-chart.tsx:25\` (\`GraphLayout\` type) | ✅ \`"force" \| "circular" \| "hierarchical"\` |
| \`app/src/components/graph-exploration-wrapper.tsx:277\` (callback signature) | ✅ |
| **\`app/src/plugins/graph/settings.ts:8\`** (Zod schema) | **❌** \`z.enum(["force", "circular"])\` |

NVL and the wrapper handle \`hierarchical\` fine, but the schema gates render. Any widget with the saved value crashes.

### 1. Add \`hierarchical\` to the schema (root fix)
\`app/src/plugins/graph/settings.ts:8\`:
\`\`\`ts
layout: z.enum(["force", "circular", "hierarchical"]).default("force"),
\`\`\`

### 2. Audit other widget schemas for missing supported values
Pattern repeats easily: lower-level chart type accepts more values than the schema gate. Grep all \`app/src/plugins/*/settings.ts\` for \`z.enum\` calls and cross-check each against the corresponding type in \`component/src/charts/*\`. Likely candidates: sunburst/treemap layout options, chart legend positions, axis tick formats. Output a checklist in the PR description so future drift is caught.

### 3. Graceful schema fallbacks (resilience)
Replace \`schema.parse(raw)\` call sites in plugin components with a defensive pattern:
- Use \`schema.safeParse(raw)\` (or per-field \`.catch(default)\`)
- On failure: log the error (\`authLogger\` or new \`pluginLogger\`), substitute defaults, render the widget anyway with a small "Settings auto-corrected" indicator in the header
[...]

Issue #916, opened 2026-06-03
[P1] Redesign dashboard import flow — unified connection mapping for NeoDash + NeoBoard


alfredo1996:
**Found during** [#895](https://github.com/alfredo1996/neoboard/issues/895) session 1a, importing a real NeoDash v2.4 dashboard.

## What's wrong with the current flow
\`app/src/app/(dashboard)/page.tsx:152-189\` has two branches with mismatched contracts:

### Branch A — NeoDash (lines 152-166)
\`\`\`ts
// NeoDash — no connection mapping needed
...
connections: {}
\`\`\`
Mapping UI is skipped. The converter (\`app/src/lib/dashboard/neodash-converter.ts:280\`) sets \`widget.connectionId = ""\` for every imported widget.

**Result**: every widget lands in the DB pointing to nothing. Every query errors at run time with "Connection not found". User got a "successful import" message with a fully broken dashboard.

### Branch B — NeoBoard (lines 167-189)
Walks \`json.connections\` and builds a mapping UI. Requires every mapping non-empty before submit.

**Bug**: when the source has e.g. a postgres connection and the target tenant has zero postgres connections, the dropdown shows zero options with no help text. User is stuck — can't satisfy \`allMapped\`, can't escape the dialog usefully.

### Cross-cutting
Import response only returns \`{ id }\`. No notes about mappings, skips, chart-type downgrades (#915), dropped params (#915), or anything else. Users find issues at runtime, not import time.

## Why one redesign, not three patches
Both branches diverge on a false premise: "NeoDash didn't expose connections, so we can skip mapping." NeoDash dashboards still **need** a connection at run time. Skipping the UI doesn't solve the problem — it makes the failure silent and delays it.

The right fix is one unified mapping step.

### 1. Unified connection extraction
\`app/src/app/(dashboard)/page.tsx\` parse step:
- **NeoBoard format**: read \`json.connections\` as today
- **NeoDash format**: synthesize a single placeholder:
  \`\`\`ts
  { 
    "neodash-default": { 
      name: \`Neo4j connection for "\${nd.title}"\`, 
      type: "neo4j" 
    } 
  }
  \`\`\`
  All widgets in a NeoDash import share this placeholder.

### 2. Unified mapping UI
Same component renders for both. Per required source connection:
- Dropdown of compatible local connections
- **No compatible target** → render: \`"No compatible <type> connections available in this tenant — create one first"\` + button linking to \`/connections\`
[...]

Issue #915, opened 2026-06-03
[P1] NeoDash importer silently drops top-level params + markdown content


alfredo1996:
**Found during** [#895](https://github.com/alfredo1996/neoboard/issues/895) session 1a, attempting to import a real NeoDash v2.4 dashboard.

## Two converter bugs, same root cause
\`app/src/lib/dashboard/neodash-converter.ts\` is incomplete and emits **no warnings** for dropped data. Two specific gaps:

### Bug A — top-level \`settings.parameters\` dropped
The converter walks \`pages\` → \`reports\` and converts queries/settings per widget. It **never reads \`nd.settings.parameters\`** — the dashboard-wide parameter map (line 240ff iterates pages; no separate handling for \`nd.settings.parameters\`).

\`convertParamSyntax\` rewrites \`$neodash_\` → \`$param_\` in queries (line 49), but the params those references point to never exist on the target dashboard. Queries silently evaluate against undefined params.

In a real-world NeoDash dashboard with 50+ params, all are lost.

### Bug B — \`text\` / \`markdown\` widget content dropped
NeoDash stores markdown body in \`report.query\`. The converter:
- Maps \`text\` / \`markdown\` → \`markdown\` chartType ✅ (\`CHART_TYPE_MAP\` line 32-33)
- Copies \`report.query\` into \`widget.query\` (line 281) ❌

But the markdown widget reads from \`settings.content\` (\`app/src/plugins/markdown/settings.ts:8\`). The imported widget renders empty.

### Bug C (cross-cutting) — silent failure mode
\`convertNeoDashWithNotes\` already returns a \`notes\` array used to flag chart-type downgrades. Neither of the two bugs above contributes to it. Users get a "successful import" message and a broken dashboard.

## Repro
\`\`\`bash
# Use the dashboard JSON from session 1a feedback (Reports Landing Page, v2.4, 50+ params, 1 text widget)
curl -X POST http://localhost:3000/api/dashboards/import -H "Content-Type: application/json" -d @reports-landing.json
\`\`\`

Expected: dashboard imported with markdown rendered + parameters carried forward (or with clear notes about what was dropped)
Actual: dashboard imported with empty markdown widget + zero parameters + no notes

### A. Top-level parameters
NeoBoard models parameters as outputs of explicit \`parameter-select\` widgets, not as a free-standing map. So a direct copy doesn't work. Hybrid:
1. Scan all widget queries for \`$param_xxx\` references
[...]

Issue #907, opened 2026-06-03
[P1] API_KEY_HMAC_SECRET treated as optional but required for the feature — dead-ends users at create time


alfredo1996:
**Found during** [#895](https://github.com/alfredo1996/neoboard/issues/895) session 1a, section E (API keys).

## Symptom
Click **Create API Key** in Settings → API Keys → red error:
> API_KEY_HMAC_SECRET is not configured. Set it in your environment variables.

The app started fine, the page loaded fine, the dialog opened fine — failure happens at click time. Nothing in setup, docs (the API-keys guide), or runtime warnings hints that the secret needs to be set before this feature works.

### 1. Setup CLI doesn't generate the secret
\`scripts/setup-local-demo.sh\` / \`neoboard demo\` generates \`ENCRYPTION_KEY\` and \`NEXTAUTH_SECRET\` automatically, but apparently not \`API_KEY_HMAC_SECRET\`. So a fresh local install has no chance of working without manual intervention.

### 2. Env-config marks it optional
\`app/src/lib/env-config.ts:61\`:
\`\`\`ts
{ key: "API_KEY_HMAC_SECRET", required: false },
\`\`\`
This is technically true (the app boots without it) but misleading: the API-keys feature is broken without it. Either:
- Mark it required (app fails to start without it; consistent with "Required for API key validation" in backup-restore.mdx)
- Or keep optional but render a startup warning + degrade the Settings > API Keys page gracefully (show "feature disabled, set API_KEY_HMAC_SECRET" empty state)

### 3. Error message is bare
\`app/src/app/api/keys/route.ts:54-57\`:
> "API_KEY_HMAC_SECRET is not configured. Set it in your environment variables."

Better:
> "API_KEY_HMAC_SECRET is not configured. Generate one with \`openssl rand -base64 32\` and add it to app/.env.local, then restart the server. Warning: rotating this invalidates all existing API keys."

Or link to docs.

### 4. User-facing docs don't mention it
- ✅ \`docs/src/content/docs/administration/backup-restore.mdx:19\` mentions it
- ✅ \`docs/src/content/docs/administration/deployment-checklist.mdx:16\` mentions it
- ❌ \`docs/src/content/docs/guides/api-keys.mdx\` (the page users land on) does NOT — goes straight from "click Create" to "use the key", omitting the prereq

## Why this all converges
The env var is treated as optional by config + setup but required by the only feature that uses it. Fix the policy mismatch and the symptoms go away.

[...]

PR #938, opened 2026-06-04
fix(dashboard): suppress self-save 'updated by' banner on revisit

### LINKED ISSUE #904: [P1] Dashboard 'updated by <self>' banner fires after own save when revisiting
**Found during** [#895](https://github.com/alfredo1996/neoboard/issues/895) session 1a.

## Repro
1. Log in as admin
2. Open a dashboard, make an edit, save
3. Navigate away
4. Come back to the same dashboard

## Expected
No banner — admin is the one who saved; the page already reflects the save.

## Actual
Banner: "Dashboard updated by admin" with a refresh prompt.

## Root cause
\`app/src/app/(dashboard)/[id]/page.tsx:148-160\`:

\`\`\`ts
if (stored === null) {
  sessionStorage.setItem(key, String(dashboardVersion));
} else if (dashboardVersion > Number(stored)) {
  sessionStorage.setItem(key, String(dashboardVersion));
  const who = dashboardUpdatedBy ?? "someone";
  setVersionBumpMsg(\`Dashboard updated by \${who}\`);
}
\`\`\`

- \`stored\` = version captured on **first** load (sessionStorage, persists across page nav)
- \`dashboardVersion\` = current server version
- The comparator only checks "server is newer than first-load snapshot"; it doesn't check whether the bump was caused by the current user

Sequence:
1. First load → stored = N
2. Self-save → server = N+1, sessionStorage still N
3. Re-fetch (next visit) → server returns N+1 > N → banner fires, reads \`updatedByName\` which is your own name

## Fix
**Primary**: update the sessionStorage key inside \`useUpdateDashboard\`'s \`onSuccess\` with the newly-returned \`version\`, so the comparator's baseline keeps up with self-saves.

**Defense-in-depth**: compare \`dashboard.updatedBy === currentUserId\` (using session) and suppress the banner if they match. Catches edge cases where the sessionStorage write races with the refetch.

## Why both
Either alone could leak the bug back if the other regresses. Together: cheap, correct.

## Acceptance
- Self-save → navigate away → return → no banner
- Other user's save → banner with their name + refresh prompt still works
- Same-user save across two tabs: refreshing tab A after tab B's save still shows the banner (consistent with prior behavior)
- E2E spec for dashboard-states / dashboard-metadata covers the self-save + revisit path

## Files
- \`app/src/app/(dashboard)/[id]/page.tsx:148-160\` — comparator + sessionStorage seed
- \`app/src/hooks/use-dashboards.ts\` — \`useUpdateDashboard\` onSuccess: write \`sessionStorage.setItem(__nb_dash_ver_<id>, version)\`
- New or extended E2E in \`app/e2e/dashboard-states.spec.ts\`

alfredo1996:
## Summary

The "Dashboard updated by X" banner was firing after a user's OWN save (and revisit), labeled with the user's own name. Caused by a stale sessionStorage baseline. Fix: update sessionStorage in `useUpdateDashboard.onSuccess` so the baseline keeps up with self-saves.

**Final PR of Phase 2.** Closes #904.

## Bug

`app/src/app/(dashboard)/[id]/page.tsx` (lines 148–160) baselines the dashboard version in sessionStorage on first load. On every subsequent render, it compares the refetched server version to the baseline. When server > baseline → banner fires.

The bug: after a successful self-save, the server bumps `version N → N+1` but the client's sessionStorage still says `N`. Refetch sees `N+1 > N` → banner triggers with the user's own name.

## Fix

In `useUpdateDashboard.onSuccess`:
```ts
if (typeof window !== "undefined") {
  const newVersion = (result as { version?: unknown } | undefined)?.version;
  if (typeof newVersion === "number") {
    sessionStorage.setItem(`__nb_dash_ver_${variables.id}`, String(newVersion));
  }
}
queryClient.invalidateQueries({ queryKey: ["dashboards", variables.id] });
queryClient.invalidateQueries({ queryKey: ["dashboards"] });
```

TanStack Query guarantees onSuccess runs before invalidateQueries' refetch fires, so the baseline is updated before the page's effect reads it.

## Defense-in-depth (NOT in this PR — drill decision)

A second-line `updatedBy === currentUserId` check was considered. Rejected because:
- Would require exposing `dashboard.updatedBy` (user UUID) in the API response — not there today
- The primary fix solves the actual race; the realistic threat model doesn't need defense-in-depth here
- Can be filed as a follow-up if a related race ever surfaces

## What still works

- **Other-user saves** still trigger the banner correctly (they don't run through this user's mutation onSuccess; the baseline stays unchanged)
- **409 conflict** path still works (onSuccess doesn't fire on save failure; baseline stays at pre-save value)
- **Two-tab same-user** edits: tab B still sees the banner when A saves — acceptable for v1 (B doesn't know A is the same person)

## Tests

**Unit** (`app/src/hooks/__tests__/use-dashboards.test.ts`) — 4 new cases on `useUpdateDashboard`:
- [x] PUT request shape (mutationFn)
- [x] onSuccess writes new version to sessionStorage
- [x] onSuccess skips write when result has no version field
- [x] onSuccess skips write when version is non-numeric

**E2E** (`app/e2e/dashboard-states.spec.ts`) — new case:
- [x] Create dashboard → save → navigate away → navigate back → assert no "Dashboard updated by" banner

**Total**: 2868/2868 unit tests pass (+4 new). Build + type-check green.

## Phase 2 — final progress

| # | Title | Status |
|---|---|---|
| #906 | SSO gating | ✅ merged |
| #916 | Import flow redesign + notes infra | ✅ merged |
| #915 | NeoDash converter fidelity | ✅ merged |
| #917 | Graph schema + safeParse resilience | ✅ merged |
| **#904** | **Self-save banner (this PR)** | 🟢 **opened** |

After this lands, **Phase 2 is complete**. Next: Phase 3 (DevEx unblockers — #907, #898, #899, #905).

## Risk

| Risk | Mitigation |
|---|---|
| sessionStorage not available SSR | `typeof window !== "undefined"` guard |
| Server response missing `version` | `typeof newVersion === "number"` check |
| Race between onSuccess and refetch | TanStack Query guarantee — onSuccess fires before invalidateQueries triggers refetch |

### DIFF
--- a/app/e2e/dashboard-states.spec.ts
+++ b/app/e2e/dashboard-states.spec.ts
@@ -65,7 +65,9 @@ test.describe("Dashboard viewer — uncovered states", () => {
     // Create a fresh dashboard
     await page.getByRole("button", { name: /New Dashboard/i }).click();
     const dialog = page.getByRole("dialog");
-    await dialog.locator("#dashboard-name").fill("Self-Save Test");
+    await dialog
+      .locator("#dashboard-name")
+      .fill(`Self-Save Test ${Date.now()}`);
     await Promise.all([
       page.waitForResponse(
         (r) =>
@@ -78,10 +80,13 @@ test.describe("Dashboard viewer — uncovered states", () => {
     ]);
     await page.waitForURL(/\/edit/, { timeout: 15_000 });
 
-    // Save (no edits required — just trigger the version bump path).
-    // The PUT bumps version from 1 → 2 server-side; the hook's onSuccess
-    // must update sessionStorage so the subsequent visit doesn't see a stale
-    // baseline.
+    // Capture the dashboard id for later assertions / cleanup.
+    const editUrl = page.url();
+    const dashboardId = editUrl.split("/").slice(-2, -1)[0];
+
+    // Save — bumps version 1 → 2 server-side. With #904's fix, the hook's
+    // onSuccess writes the new version to sessionStorage so the subsequent
+    // view-mode load sees a fresh baseline and doesn't fire the banner.
     await Promise.all([
       page.waitForResponse(
         (r) =>
@@ -93,21 +98,22 @@ test.describe("Dashboard viewer — uncovered states", () => {
       page.getByRole("button", { name: "Save" }).click(),
     ]);
 
-    // Leave dashboard (back to list)
+    // Leave edit → view mode (the route where the version-bump effect runs)
     await page.getByRole("button", { name: /Back/ }).click();
-    await page.waitForURL(/\/dashboards/, { timeout: 10_000 });
-
-    // Re-open the same dashboard
-    await page.getByText("Self-Save Test", { exact: true }).click();
+    // Back goes to /<id>, not /dashboards
     await page.waitForURL(/\/[\w-]+$/, { timeout: 10_000 });
 
-    // Wait briefly for any version-bump effect to settle. The banner would
-    // appear in the same paint as the dashboard content if the bug were
-    // present, so a short stability window is sufficient.
+    // Wait briefly for the version-bump effect to settle. Banner would
+    // appear in the same paint if the bug were present.
     await page.waitForTimeout(500);
 
     // Critical assertion: NO "Dashboard updated by" banner
     await expect(page.getByText(/Dashboard updated by/i)).toHaveCount(0);
+
+    // Clean up to avoid polluting other tests
+    if (dashboardId) {
+      await page.request.delete(`/api/dashboards/${dashboardId}`);
+    }
   });
 });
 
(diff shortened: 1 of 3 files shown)


alfredo1996:
### Fixed E2E test (commit e4209b81)

E2E shard 2/5 failed on my new test because I had the wrong URL pattern after clicking "Back" — Back goes to /<id> (view mode), not /dashboards (list). Simplified the test: edit → save → Back (view mode) is exactly where the version-bump effect runs, so the simpler flow exercises the bug directly without a roundtrip through the list.

Also added a per-test timestamp suffix to the dashboard name + cleanup at end to avoid leaking dashboards into other tests (which may have caused the secondary `dashboards.spec.ts:106` flake on the same run).

Watching CI on e4209b81.

Issue #1052, opened 2026-06-12
[P2] Graph-heavy dashboards exhaust WebGL contexts — "Too many active WebGL contexts" / context-loss warnings


alfredo1996:
Found during v1.1 dogfood review session 4 (#895), lens: regression/performance.

## Problem
On the seeded "Chart Reference" dashboard (many NVL graph widgets across its pages/scroll), the console emits repeated `Too many active WebGL contexts. Oldest context will be lost.` and `WebGL context lost - visualization will stop working!` (3× each observed). Browsers cap simultaneous WebGL contexts (~16); each NVL graph holds one, so a dashboard with several graph widgets evicts older contexts and those graphs stop rendering.

A SINGLE graph widget shows NO such warning (verified in the same session), so this is specifically a resource-exhaustion issue on graph-dense dashboards, not a per-widget bug.

## Fix options
- Only mount/initialize NVL for graph widgets currently in the viewport (intersection-observer based lazy mount/unmount), releasing the WebGL context when scrolled away.
- Or cap concurrently-rendered graph widgets and show a "scroll to render" placeholder for the rest.
- Ensure context-loss is handled gracefully (re-acquire on scroll-back) rather than leaving a dead canvas.

## Required tests
- **E2E (Playwright):** load a dashboard with N (>WebGL cap) graph widgets → assert no "context lost" console errors and that scrolling a graph into view renders it. (Heavy-widgets spec is the right home.)
- **Component:** graph wrapper releases its WebGL context on unmount/scroll-out (mock NVL, assert dispose called).

Part of #895 (session 4).

alfredo1996:
Fixed in #1074, merged to `release/1.1`.

PR #1061, opened 2026-06-12
fix(app): apply table styling rules with target "color" as text color

### LINKED ISSUE #1057: [P1] Rule-Based Styling: table rules with target "color" silently apply no text color
Found during v1.1 dogfood review session 7 (#895), lens: regression/design. Root-caused. The seeded "Rule-Based Styling" showcase is visibly affected.

## Symptom
On a table widget, a conditional-styling rule whose target is text color does NOT colour the text. The seeded "Rule-Based Styling" dashboard shows threshold rules applying **bold** but not the intended green/amber/red **text colour**.

## Root cause — target-vocabulary mismatch
- The table plugin declares its styling targets as `backgroundColor` / `textColor` (`app/src/plugins/table/component.tsx:58`).
- The renderer only honours those two (`app/src/components/table-renderer.tsx:184-191`): `if (target === "backgroundColor") … if (target === "textColor") …`.
- But the canonical text-colour target value elsewhere is **`"color"`**: the seeded showcase data, `app/src/lib/dashboard/migrate-color-thresholds.ts:39` (`target: "color"`), the import schema (`dashboard-import.ts:17` enum `["color","backgroundColor","textColor"]`), and EVERY other chart plugin (single-value, pie, line, sankey, gauge, map, graph, …) all use `value: "color"` for text colour.
- So a table rule with `target: "color"` matches neither renderer branch → no colour applied. The independent `rule.bold` branch still fires, which is exactly why bold shows but colour doesn't.

## Scope
- Affects seeded/migrated/imported dashboards and any rule copied from another chart type (all use `"color"`).
- New table rules authored via the table editor use `"textColor"` (the table plugin's own target) and DO work — so it's a partial break, but it breaks the out-of-the-box showcase and silent-fails on import/migration.

## Fix
- In `table-renderer.tsx`, treat `target === "color"` as text colour (alias of `textColor`): `if (target === "textColor" || target === "color") { style.color = color }`. Also update the `!stylingRules.some(r => r.target === "textColor")` auto-contrast guard to include `"color"`.
- Better: unify the target vocabulary so text colour has ONE canonical value across plugins (prefer `"color"` since it's the majority), and migrate the table plugin + seed/migrate accordingly.

## Required tests
- **Unit (app):** table-renderer style resolver — a rule `{target:"color", color:"#22c55e"}` that matches → produces `style.color === "#22c55e"`. Pin both `"color"` and `"textColor"`.
- **Unit (component):** styling-rule resolver vocabulary — assert `"color"` and `"textColor"` are equivalent for tables (whichever direction the fix unifies).
- **E2E (Playwright):** open the seeded "Rule-Based Styling" dashboard → assert a cell over the green threshold has the expected computed text colour, not just bold.

Part of #895 (session 7).

alfredo1996:
## Summary
Fixes #1057 — on table widgets, conditional-styling rules whose target is text colour silently applied nothing. The seeded "Rule-Based Styling" showcase showed bold but no colour.

## Root cause
Two spellings exist for "colour the text": the table plugin editor emits `"textColor"`, but the seed data, `migrate-color-thresholds.ts`, dashboard import, and every other chart plugin emit `"color"`. `table-renderer.tsx` only handled `"textColor"`/`"backgroundColor"`, so a `"color"` rule matched neither branch. The independent `bold` flag still fired — exactly the reported symptom.

## Fix
- Extract the inline row-style closure into a pure, unit-tested helper `resolveStylingRuleRowStyle` (`app/src/lib/widget/table-styling.ts`).
- Treat `"color"` as an alias of `"textColor"` (a shared `TEXT_COLOR_TARGETS` set), including the auto-contrast guard so an explicit text-colour rule is never overridden by the contrast fallback.

## Tests (TDD — red→green verified)
- New `app/src/lib/widget/__tests__/table-styling.test.tsx` (8 cases): `"color"` and `"textColor"` both yield text colour, background + bold paths, contrast auto-fill, explicit-colour-not-overridden, no-match → undefined, defaultCol fallback. Confirmed the suite fails (4×) if the `"color"` alias is removed.
- Component styling-rule suite green (33); component package green (1737).
- Type-check + ESLint clean.

## E2E
Ran `styling-rules`, `grid`, `table-column-filters`, `charts` specs (Docker destroyed first): **49 passed**, styling-rules included. 2 unrelated failures in NVL **graph** chart exploration (`charts.spec.ts`) asserting "no console errors" — these are the WebGL context-loss flakes tracked in #1052, on a surface untouched by this PR (table-only diff). CI will re-run the full suite.

### DIFF
--- a/app/src/components/table-renderer.tsx
+++ b/app/src/components/table-renderer.tsx
@@ -9,13 +9,13 @@ import {
   DataGridPagination,
   parseColorThresholds,
   resolveThresholdColor,
-  resolveStylingRuleColor,
   interpolateColor,
   contrastTextColor,
 } from "@neoboard/components";
 import type { StylingRule, ColorScaleConfig } from "@neoboard/components";
 import type { ColumnDef } from "@tanstack/react-table";
 import { parseGroupByColumns } from "@/lib/widget/table-utils";
+import { resolveStylingRuleRowStyle } from "@/lib/widget/table-styling";
 
 const AGG_SYMBOLS: Record<string, string> = {
   sum: "Σ",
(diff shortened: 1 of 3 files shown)


alfredo1996:
@coderabbitai review

Issue #969, opened 2026-06-10
[P1] cli: demo prints ready banner + admin credentials after a silently skipped seed


alfredo1996:
**Found during** the 2026-06-10 full package audit on `release/1.1` (report: `claude_code_docs/review-2026-06-package-audit.md`). Related: #879.

## Problem

`scripts/seed-demo.mjs:167-184` exits 0 with "skipping seed" when `app/.env.local` is missing. The CLI then reports success: `spinner.succeed("PostgreSQL seeded with demo data")` (`cli/src/commands/db/seed.ts:80`) and the banner advertises `admin@neoboard.local / admin123` (`cli/src/commands/demo.ts:26-33`) — **a user that was never created**.

This is the *default* fresh-clone Docker path, because `runInit` only generates `.env.local` in local mode (`init.ts:28-43`) and `runEnv` tells Docker users it's "Not needed" (`env.ts:81-86`).

Related second-order bug: when `.env.local` *does* exist, the host-side seed encrypts connector configs with the host `ENCRYPTION_KEY` while the app container runs with the compose-baked key — seeded connectors can't be decrypted at runtime.

## Suggested fix

- `seed-demo.mjs` should exit non-zero (or the CLI should treat "skipped" as a distinct status) so `demo` fails loudly instead of printing false credentials.
- Decide on one key source for Docker demo mode (inject the host-generated key into compose via env) so seeding and runtime agree.

alfredo1996:
Fixed in #1032 (merged to release/1.1): seed-demo exits non-zero on missing secrets (no more fake-creds banner); Docker mode threads docker/.env ENCRYPTION_KEY + localhost DSN so host seed and container agree. Verified live: exit 1 with 'admin was NOT created'.

PR #1062, opened 2026-06-12
fix(cli): seed connectors with the docker/.env encryption key on every path

### LINKED ISSUE #1039: [P1] CLI: demo seed / db seed / db reset encrypt connector credentials with the wrong key in docker mode
Found during v1.1 dogfood review session 2 (#895) — this silently broke the entire demo environment mid-review.

## Symptom
After `neoboard setup` + `neoboard demo seed`, every seeded connection fails its test with HTTP 500. Server log: `Error: Unsupported state or unable to authenticate data` from `Decipheriv.final` — AES-GCM decrypt failure. All demo dashboards are broken with no actionable error anywhere.

## Root cause
The #969 fix introduced `readDockerEnvSecrets()` so host-side seeding encrypts with the SAME `ENCRYPTION_KEY` the app container uses (from `docker/.env`). But only `neoboard demo` (full command, via `runDbSeed({dockerNetwork: true})`) got the fix. Two other entry points missed it:

1. **`neoboard demo seed`** — `runDemoSeed()` spawns `seed-demo.mjs` with `dockerEnv()` (`cli/src/commands/demo.ts:94`) which only injects `NEO4J_HOST`/`PG_HOST`. No `ENCRYPTION_KEY`/`DATABASE_URL` → seed-demo.mjs falls back to `app/.env.local`, which can hold a stale/divergent key (on dev machines it holds the E2E test key). Connections are then encrypted with the wrong key.
2. **`neoboard db seed` / `neoboard db reset`** — `cli/src/index.ts:287` calls `runDbSeed({neo4j, demo})` without `dockerNetwork`, and `seedPostgres()` only merges docker secrets when `dockerNetwork === true` (`cli/src/commands/db/seed.ts:69-90`). In docker mode this both uses the wrong key AND writes `localhost` connection URIs that don't resolve inside the container network.

## Fix
- Pass docker-mode env (secrets + internal hostnames) from ALL seed entry points — extract one shared "seed env for current mode" helper used by `runDemoSeed`, `runDbSeed`, `runDemo`.
- `db seed`/`db reset` should derive `dockerNetwork` from `getMode()` instead of a caller flag that callers forget.
- Consider: `seed-demo.mjs` could verify round-trip decryptability against the app (or at least print which key source it used) so a mismatch fails loudly instead of producing silently broken connectors.

## Repro
On a machine where `app/.env.local` exists with any ENCRYPTION_KEY different from `docker/.env`:
`neoboard setup` → `neoboard demo seed` → open /connections → all seeded connections show Error; server logs decrypt failures.

Part of #895 (session 2). Related: #969, #1040-style UX of the failure is filed separately.

## Required tests
- **Unit (vitest, cli):** `runDemoSeed` in docker mode — asserts the spawned seed process env contains `ENCRYPTION_KEY`/`DATABASE_URL` from `readDockerEnvSecrets()` (mock `run`/`readDockerEnvSecrets`, assert merge). Mirror the existing test for `runDbSeed`'s dockerNetwork branch.
- **Unit (vitest, cli):** `db seed` / `db reset` command wiring — asserts `dockerNetwork` derives from `getMode() === "docker"` rather than defaulting to false.
- **Unit (vitest, cli):** precedence test for the shared seed-env helper: process env > docker/.env > app/.env.local, and loud failure (non-zero exit) when no key source exists.
- **Round-trip guard:** test that a connection seeded via the helper's key decrypts with the same `docker/.env` key (encrypt/decrypt round-trip in the seed path), pinning the actual failure mode found in review.

alfredo1996:
## Summary
Fixes #1039 — in Docker mode, `neoboard demo seed`, `db seed`, and `db reset` encrypted seeded connector credentials with the **wrong** `ENCRYPTION_KEY`, so every seeded connection failed to decrypt at runtime (`Unsupported state or unable to authenticate data` → HTTP 500). This silently broke the demo/onboarding experience and took down the dogfood review environment mid-run.

## Root cause
The #969 fix (use the `docker/.env` key for host-side seeding) only reached `neoboard demo` (full flow), which passed `dockerNetwork: true`. The other seed entry points missed it:
- `runDemoSeed` used a local `dockerEnv()` that injected only `NEO4J_HOST`/`PG_HOST`, not `ENCRYPTION_KEY`/`DATABASE_URL`.
- `db seed` / `db reset` called `runDbSeed` without `dockerNetwork`, so `seedPostgres(false)` skipped the docker secrets entirely and wrote `localhost` URIs.

## Fix
- New shared `buildSeedEnv(config)` in `lib/docker-env.ts`, keyed on `getMode()`: in Docker mode it threads internal hostnames, a localhost DSN, and the `docker/.env` `ENCRYPTION_KEY`; in local mode it returns `process.env` unchanged.
- All seed paths now use it. Removed the per-call `dockerNetwork` flag (the footgun callers forgot) and the duplicate `dockerEnv()`.

## Tests (TDD — red→green)
- `buildSeedEnv` unit tests (docker/local mode, no-override of caller-set values).
- `seedPostgres`/`runDbSeed` assert delegation to `buildSeedEnv` (regression guard: `seedPostgres.length === 0`).
- `runDemoSeed` asserts it seeds with the `buildSeedEnv` environment.
- Full CLI suite green (283 passed); CLI builds clean.

## End-to-end verification
Ran the previously-broken `neoboard db seed --demo` in docker mode → all 4 seeded connections decrypt with the `docker/.env` key and use docker-internal hostnames (`neoboard-postgres`, `neoboard-neo4j`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved Docker Compose environment variable configuration issues preventing proper value interpolation during container stack deployment.
  * Improved consistency of database configuration and secrets handling across Docker and local operating modes.

* **Refactor**
  * Simplified database seeding environment initialization logic.

* **Chores**
  * Enhanced CI/CD workflow Docker environment setup procedures.

### DIFF
--- a/.github/workflows/cli-integration.yml
+++ b/.github/workflows/cli-integration.yml
@@ -49,7 +49,7 @@ jobs:
         run: npm -w connection run build
         continue-on-error: true
 
-      - name: Generate app/.env.local
+      - name: Generate env files (app/.env.local + docker/.env)
         run: |
           cat > app/.env.local <<'ENVEOF'
           DATABASE_URL=postgresql://neoboard:neoboard@localhost:5432/neoboard
@@ -60,6 +60,18 @@ jobs:
           ENVEOF
           # Remove leading whitespace from heredoc
           sed -i 's/^          //' app/.env.local
+          # docker-compose.full.yml interpolates these from docker/.env (the
+          # file `neoboard start`/`demo` generates). CI starts the stack with
+          # raw `docker compose up`, so generate it here with the SAME secrets
+          # as app/.env.local — otherwise compose fails on the required
+          # ${API_KEY_HMAC_SECRET:?} interpolation and seeded connectors would
+          # be encrypted with a divergent key (#1039).
+          cat > docker/.env <<'ENVEOF'
+          ENCRYPTION_KEY=b8c0dbaad415694973d7cf4a3a40d4e53fc940493a6362ecff4dae45245e05d9
+          NEXTAUTH_SECRET=d0eece19938fc5e2e3e45ed76fb5c92b0fc6ba2c4f213404ccca7ea0e641cd65
+          API_KEY_HMAC_SECRET=7f3b9c2a5e8d4f1b0a6c9e8d7f2a4b1c5e8d7f2a4b1c5e8d7f2a4b1c5e8d7f2a
+          ENVEOF
+          sed -i 's/^          //' docker/.env
 
       - name: Build Docker image
         run: docker build -t neoboard:integration-test .
--- a/cli/src/__tests__/commands/db/seed.test.ts
+++ b/cli/src/__tests__/commands/db/seed.test.ts
@@ -15,6 +15,7 @@ vi.mock("../../../lib/docker.js", () => ({
 
 vi.mock("../../../lib/config.js", () => ({
   paths: { root: "/project" },
+  getMode: vi.fn(() => "docker"),
   readProjectConfig: vi.fn(() => ({
     neo4j: { user: "neo4j", password: "neoboard123" },
     postgres: { user: "neoboard", password: "neoboard", database: "neoboard" },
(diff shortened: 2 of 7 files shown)


alfredo1996:
Follow-up commit (db59116d) bundles a CI fix: the **CLI Integration Tests** workflow was deterministically broken (pre-existing) — `docker-compose.full.yml` interpolates `API_KEY_HMAC_SECRET`/`ENCRYPTION_KEY`/`NEXTAUTH_SECRET` as required vars from `docker/.env`, but the workflow only wrote `app/.env.local`, so `docker compose up` failed before any seed/test step ran. It surfaced now because this is the first CLI PR to run the workflow since the `${VAR:?}` markers landed.

The fix generates `docker/.env` with the same secrets, so the stack starts and the seeded key matches the app container — the exact invariant #1039 is about. Verified locally: `docker compose -f docker/docker-compose.full.yml config` now interpolates cleanly.

Issue #922, opened 2026-06-03
[P3] Stop hook fires haiku checklist on every turn — make periodic or smarter


alfredo1996:
**Found during** [#895](https://github.com/alfredo1996/neoboard/issues/895) pre-polish review.

## Problem
`.claude/settings.json` defines a `Stop` hook that fires a `claude-haiku-4-5-20251001` prompt on EVERY assistant response. The prompt checks: tests run? linting? screenshots? Useful intent, but firing per-turn means an extra LLM call on every reply, adding latency + cost. The hook also frequently responds with allow for planning turns where nothing was edited.

## Proposed
Make the hook smarter so it only fires when there's something to check:
- Skip if no Edit/Write tool was used since the previous turn
- Or fire only at end-of-task markers (user says "done", "ship it", etc.)
- Or fire only after Bash git commit is invoked (caught by enforce-e2e.sh, repurpose pattern)

Simplest viable: a one-line shell guard before the prompt fires that checks for tracked edits since last stop. If none, skip.

## Acceptance
- Stop hook fires only when code was edited that turn
- Per-turn cost reduced
- E2E and lint enforcement still works (existing PreToolUse hooks handle the heavy lifting)

alfredo1996:
Closing as part of the v1.5 release plan (2026-09-10).

Fixed: .claude/hooks/stop-check.sh replaced the per-turn Haiku prompt hook and cites #922.

Checked against `origin/release/1.5`: .claude/hooks/stop-check.sh on origin/release/1.5 says it 'Replaces a prompt-type hook that invoked Haiku on EVERY turn-completion ... (#922)'. settings.json now wires Stop to that command hook (5s timeout), which blocks only when the .e2e-needed marker exists. Commit b8d88dec: 'replace per-turn Stop haiku'.

PR #1065, opened 2026-06-12
chore(agents): cut browser-agent token waste; replace per-turn Stop haiku

alfredo1996:
## Why
Token-efficiency review of the agent/skill/hook config, driven by waste observed during the v1.1 dogfood review.

## Browser agents (feature-reviewer, ux-crawler, user-sim-admin, user-sim-creator)
Two failure modes burned tokens repeatedly:
1. **"Screenshot at EVERY step/page"** — an image token on every interaction, which routinely exhausted the agents' turn budget mid-run.
2. **Findings only in the final message** — when an agent hit its `maxTurns`, the entire run was lost; I had to recover findings from raw transcripts and re-run sessions.

Each agent now:
- **Persists findings to a disk file as it goes** — the file is the deliverable, so a turn-limit death loses nothing; the final message is a short summary + path.
- **Prefers the cheap text `snapshot` over the image `screenshot`** — screenshots become evidence-for-a-finding, not narration.
- **Batches inspection** and knows refs go stale between snapshots.
- **Carries the NeoBoard browser gotchas** (dialog scoping, CodeMirror query entry, `neoboard_demo_public` schema, seeded logins) so it stops rediscovering them every session.

## Stop hook (#922)
Was a `prompt`-type hook invoking **Haiku on every turn-completion**, re-reading the growing transcript — a compounding per-turn cost for little signal (lint already auto-runs on each edit; the real gate, "UI changed without Playwright," is tracked by the `.e2e-needed` marker). Replaced with a **deterministic command hook** (`stop-check.sh`) that reuses that marker: **zero model cost**, and more reliable than re-deriving the fact from the transcript. Tested across all three states (no marker → allow, marker → block, `stop_hook_active` → allow).

## CLAUDE.md
Dropped the redundant "Playwright CLI (for browser agents)" subsection — that detail now lives in the agent definitions (and shouldn't be in the main-loop instructions, which are loaded every session + injected into every subagent).

## Not done here (proposed, needs your call)
- Further CLAUDE.md trimming (the agent-pipeline prose partly duplicates agent descriptions) — left intact to avoid unilaterally cutting project instructions.
- `#897` testcontainers v12 bump still needs a real migration (separate).

### DIFF
--- a/.claude/agents/feature-reviewer.md
+++ b/.claude/agents/feature-reviewer.md
@@ -12,6 +12,23 @@ maxTurns: 80
 
 You are a QA engineer reviewing a specific feature in the NeoBoard web application running at **http://localhost:3000**.
 
+## Token discipline & durability — READ FIRST
+
+You run on a hard turn budget (`maxTurns`). Browser calls are the expensive part, so spend them on coverage, not narration:
+
+- **Persist findings to disk as you go.** The moment you confirm a finding or finish a flow, append it to a findings file via Bash (`mkdir -p claude_code_docs/<task>/ && cat >> claude_code_docs/<task>/findings.md`). The file IS your deliverable — if you hit the turn limit mid-run, nothing is lost. Your final message is a short verdict + the file path, not the full report.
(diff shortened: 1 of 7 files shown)


alfredo1996:
Closes #922 (the Stop-hook-fires-haiku-every-turn issue) — the deterministic `stop-check.sh` replaces it.

Issue #978, opened 2026-06-10
[P2] connection: row limiting buffers entire result in memory; documented p-queue/AbortSignal/HKDF invariants don't exist


alfredo1996:
**Found during** the 2026-06-10 full package audit on `release/1.1` (report: `claude_code_docs/review-2026-06-package-audit.md`).

## Problem — documented safety invariants that don't exist in the code

CLAUDE.md states row limits use "cursor/stream consumption with MAX_ROWS+1 pattern". Reality: both connectors **buffer the entire result set in memory** then slice — Neo4j materializes all `res.records` then `result.slice(0, config.rowLimit)` (`connection/src/neo4j/Neo4jConnectionModule.ts:94,106-112`, with a TODO admitting it at line 105); pg `client.query` buffers all rows then slices (`connection/src/postgresql/PostgresConnectionModule.ts:124,141-143`). A query matching 10M rows loads 10M rows into Node before truncating.

Also documented but **absent from the entire repo** (verified by grep):
- per-connector `p-queue` concurrency — no such dependency or usage anywhere
- AbortSignal-based pg timeouts — pg uses `SET LOCAL statement_timeout` (fine, but the doc is wrong)
- "HKDF-SHA256 envelope" crypto — `app/src/lib/crypto/crypto.ts:6-14` uses the raw env key directly with AES-256-GCM (sound, but no HKDF, no envelope)

## Suggested fix

Two-part:
1. **Implement** streaming row limits (pg-query-stream / Neo4j `subscribe`) — the real memory-safety fix — and decide whether per-connector queueing is actually wanted.
2. **Reconcile CLAUDE.md** (and ARCHITECTURE.md) so guardrails, reviewers, and agents check against reality instead of fiction.

alfredo1996:
## Blocked on a dependency decision (overnight run, 2026-06-11)

The charter choice was **full streaming** (pg cursor reads rowLimit+1 then stops; Neo4j subscribe). Implementation surfaced a gate I shouldn't cross autonomously:

- **PostgreSQL**: `pg` core has no built-in row-by-row streaming. True cursor consumption needs **`pg-query-stream`** (which wraps `pg-cursor`) — a new production dependency, not currently in the lockfile. Adding it is blocked by the dependency-install guard and is a call for the maintainer (bundle size, an additional supported surface).
- **Neo4j**: the driver's reactive/async-iterable API supports this with **no new dependency** — doable now.

[...]

PR #1086, opened 2026-06-14
docs(connection): reconcile documented invariants with reality (#978)

alfredo1996:
Part of #978 (the doc-reconciliation half; the streaming row-limit fix follows in a separate PR).

The docs claimed connector/crypto invariants the code doesn't implement, so guardrails, the `code-reviewer` agent, and contributors were checking against fiction. Corrected everywhere they appear — CLAUDE.md, ARCHITECTURE.md, `docs/APP_IMPLEMENTATION_GUIDE.md`, `.github/SECURITY.md`, `.claude/agents/code-reviewer.md`, `.claude/skills/harden/SKILL.md`:

- **Concurrency** — there is no per-connector `p-queue`. The drivers' own connection pools (node-pg `Pool`, Neo4j driver pool) handle it.
- **Timeouts** — PostgreSQL uses `SET LOCAL statement_timeout` (not `AbortSignal`); Neo4j uses the managed-transaction `timeout`.
- **Crypto** — AES-256-GCM with the raw 32-byte `ENCRYPTION_KEY` used directly as the key (no HKDF derivation, no envelope/data-key wrapping); ciphertext is `iv:authTag:ciphertext` (base64); rotation via `ENCRYPTION_KEY_OLD`. **No crypto code change** — changing the scheme would break existing ciphertext and the lost-key-unrecoverable contract.

The row-limit *"cursor/stream … MAX_ROWS+1"* claim is intentionally **left as-is** — the follow-up PR implements streaming (the real memory fix) and makes that line true, so each PR's docs match its own reality (no churn).

**Docs-only — no code or behavior change.** ASCII diagrams/tables re-verified for alignment; markdown reflowed by prettier.

## Summary by CodeRabbit

* **Documentation**
  * Updated query timeout handling documentation to reflect driver-level transaction timeouts for database operations.
  * Revised credential encryption documentation to clarify the encryption scheme and key rotation support.
  * Updated connection pool saturation handling guidance in hardening guidelines.
  * Clarified concurrency model documentation to reflect driver-managed connection pools.

* **Chores**
  * Updated agent instructions and architectural guidelines for consistency with new security and performance policies.

### DIFF
--- a/.claude/agents/code-reviewer.md
+++ b/.claude/agents/code-reviewer.md
@@ -39,7 +39,7 @@ Senior reviewer for NeoBoard. Check staged/unstaged changes against rules, then 
 
 - Read-only transactions for non-Form widgets (PostgreSQL: `BEGIN READ ONLY`, Neo4j: session access mode)
 - Row limits use MAX_ROWS+1 pattern, never LIMIT on user queries
-- Timeouts at driver level (AbortSignal for pg, native for Neo4j)
+- Timeouts at the driver/transaction level (PostgreSQL: `SET LOCAL statement_timeout`; Neo4j: managed-transaction timeout)
 - User queries never modified or wrapped
 
 ### Architecture (HIGH)
--- a/.claude/skills/harden/SKILL.md
+++ b/.claude/skills/harden/SKILL.md
@@ -30,7 +30,7 @@ Test with extreme inputs by reading code and identifying vulnerabilities:
 - **Connection timeout**: 30s timeout hit — clear error message with retry?
 - **Auth failure**: Invalid credentials — redirect to connection settings, not cryptic error?
 - **Connection lost mid-query**: WebSocket/driver disconnect — widget error state with retry?
-- **Rate limiting**: p-queue saturation — queued indicator or backpressure feedback?
+- **Connection-pool saturation**: driver pool exhausted under load — queued/backpressure feedback rather than a hang?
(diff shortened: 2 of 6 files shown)


alfredo1996:
Addressed CodeRabbit review:

**Concurrency (Major — guide self-contradiction):** I verified the code rather than trusting the issue's premise. There **is** a real per-connector concurrency queue — a bespoke `QueryScheduler` (`app/src/lib/query/scheduler.ts` + `scheduler-registry.ts`, one per connectionId) with priority tiers (1=interactive > 2=load > 3=refresh), per-user round-robin fairness, `maxConcurrent`/`maxPerUser` caps, backpressure (queue-full → 503), queue timeouts, and P3 load-shedding, all tuned via `QUERY_*` env vars. The only accurate part of the original critique was that it is **not** the `p-queue` npm package. Rewrote every concurrency claim (CLAUDE.md, APP_IMPLEMENTATION_GUIDE.md prose + table, harden SKILL.md) to describe the scheduler accurately, resolving the contradiction with the guide's own scheduler section.

**Key format (Minor):** Reworded `ENCRYPTION_KEY` as 'a 64-character hex string = 32 bytes' to match the parser's `hex.length === 64` check, across CLAUDE.md, ARCHITECTURE.md, SECURITY.md, and APP_IMPLEMENTATION_GUIDE.md.

Row-limit MAX_ROWS+1 streaming claim left untouched here — PR 2 implements streaming so the claim becomes true (per the split agreed in the drill).

PR #1269, opened 2026-07-26
fix(ci): trigger on scripts/, and actually run its 71 orphaned tests (#1263)

### LINKED ISSUE #1263: fix(ci): scripts/ and eslint.config.js can land without lint running
## Problem

`ci.yml` triggers on:

```yaml
paths:
  - 'app/**'
  - 'component/**'
  - 'connection/**'
  - 'cli/**'
  - 'sonar-project.properties'
  - 'Dockerfile'
  - '.github/workflows/ci.yml'
  - 'package.json'
  - 'package-lock.json'
```

But the lint job runs `eslint .` from the repo root, which **does** cover `scripts/`, `eslint.config.js`, and other root-level JS. So a change confined to those paths skips CI entirely — including lint — while still being lintable code.

Observed on #1262: it added `scripts/shoot-stories.mjs` and only two checks ran (submodule rejection, CodeRabbit). GitHub reported `CLEAN` because no *required* check was missing. I linted the file locally before merging, but that was discipline, not a gate.

## Why it matters

`scripts/` is not incidental — it holds `seed-demo.mjs`, the plugin/connector import generators that run as `predev`/`prebuild`, `validate-seed-data.ts`, and now `shoot-stories.mjs`. A syntax error in a generator breaks every dev and build. Nothing currently catches it on the PR.

Note this is the mirror image of #1255: there CI fails for reasons unrelated to the change; here CI stays silent when it should speak.

## Proposed fix

Add the paths the root lint actually covers:

```yaml
  - 'scripts/**'
  - 'eslint.config.js'
  - 'tsconfig*.json'
```

Prefer being explicit over switching to a broad trigger — the path filter exists so docs-only changes skip the full matrix (#975), and that benefit is worth keeping.

Consider also whether the whole `unit-tests` matrix needs to run for a `scripts/` change, or only `lint` + the `scripts/__tests__` suite. Those tests exist (`scripts/__tests__/*.test.mjs`) but no workflow appears to invoke them — worth confirming, and if true it is the more interesting half of this issue.

## Required tests

- open a PR touching **only** `scripts/` and confirm the lint job now runs
- introduce a deliberate lint error in a `scripts/` file and confirm the PR **fails** — the point is that it now catches something
- confirm a docs-only change still skips the matrix, so #975's benefit is preserved
- determine whether `scripts/__tests__` is executed by any workflow; if not, wire it in or say plainly that it is local-only

## Acceptance criteria

- [ ] `scripts/**` and root config files trigger lint
- [ ] Demonstrated to fail on a deliberate error
- [ ] Docs-only changes still skip the full matrix
- [ ] `scripts/__tests__` execution status resolved either way

alfredo1996:
Closes #1263 — and corrects its premise.

## Two gaps; the second was the real one

CI's path filter omitted `scripts/`, so a change confined to it **skipped CI entirely**. That directory holds the plugin and connector import generators that run as `predev`/`prebuild` — a break there stops every dev and every build.

More seriously: **`scripts/__tests__` was run by no workflow at all.** 71 tests, including coverage of those generators, had never gated a PR.

## Why one command didn't already exist

The directory mixes two runners — three files use vitest, two use Node's built-in `node:test`. A single `vitest run scripts/__tests__` fails on the latter with `No test suite found in file`.

I chose **not** to convert them. Those two files currently **pass** (36 assertions), and converting 59 `assert.*` calls — 11 of them multi-line — via regex risks quiet damage to green tests for no functional gain. Instead the two runners' file sets are now **self-maintaining by naming**:

- `*.test.mjs` → vitest
- `*.node-test.mjs` → `node --test`

Adding a file to either set needs no script change. Standardising on one runner is worth doing, but not as a drive-by rewrite.

## Correction to my own issue

I wrote that `eslint .` covers `scripts/`. **It does not.** `eslint.config.js` has a config block only for `**/*.{ts,tsx}`, so root-level JS matches no block and is **silently unlinted** — enabling it surfaces 33 problems across 8 files. That is a separate change, filed rather than smuggled in here.

Adding `scripts/**` to the paths still matters, because it's what makes the new `test:scripts` step run.

## Also fixed

Vitest was discovering stale agent worktrees under `.claude/worktrees/` as duplicate test files; now excluded.

## Verification

- `npm run test:scripts` — 3 vitest files / 35 tests, plus 36 `node:test` assertions
- `npm run verify` now includes it: 248 + 115 + 28 + 3 vitest files, all green
- Attempted to prove lint catches a `scripts/` error and it **did not** — which is exactly how the eslint-coverage gap above was discovered

## Summary by CodeRabbit

- **Tests**
  - Expanded automated coverage for plugin import generation, manifest validation, and enterprise setup workflows.
  - Added checks for command-line help, dry-run behavior, error handling, and generated output consistency.

- **Chores**
  - Updated verification and continuous integration checks to include script-related tests.
  - Ensured changes to scripts and lint configuration are appropriately validated in CI.


alfredo1996:
@coderabbitai review

PR #1271, opened 2026-07-26
fix(charts): one gridline treatment for cartesian axes, keep category labels when compact (#1247)

### LINKED ISSUE #1247: fix(charts): grid weight and axis labels are inconsistent between chart types
## Problem

Two issues that make charts on one dashboard look like they came from different tools:

1. **Gridline weight differs by chart type.** The standalone bar chart draws heavy, near-solid horizontal gridlines; the line chart's are almost invisible. Evidence: `_assets/styling-review/barchart-light.png` vs the line chart in `appshell-light.png`.
2. **The bar widget renders with no axis labels at all** in the dashboard context — four bars, no category names, no value axis. Unlabelled bars are not readable data.

§4 of the design-review skill specifies shared chart defaults (grid insets, bottom legend, tooltip) but says nothing about gridline weight, so each chart module drifted.

## Proposed fix

- One gridline weight/opacity for all cartesian charts, expressed as a token (and different per theme, consistent with how chart colours already work).
- The category axis must survive small containers. `BaseChart` already has a documented compact mode (below 300px wide: hide value labels, tighten grid) — the failure is that it currently degrades to *nothing* rather than to a readable minimum. Degrade to a legend or truncated labels instead of dropping identification entirely.

## Required tests

**Component**
- both bar and line charts resolve the same gridline style token
- compact mode retains some category identification (labels or legend); assert it is not empty
- snapshot/assert the ECharts option object for grid + axis config across chart types, so drift fails loudly

**Visual**
- regenerate the bar/line screenshots in both themes and compare gridline weight side by side

## Acceptance criteria

- [ ] Shared gridline treatment across cartesian charts, theme-aware
- [ ] Bar widget never renders without category identification
- [ ] Option-shape tests prevent future per-chart drift

alfredo1996:
Closes #1247

## What was actually wrong

The issue reported "bar draws heavy gridlines, line's are almost invisible". Reading the code, bar and line **both** already inherit the theme's `splitLine` — the drift was elsewhere:

- **`timeAxis` was never registered in either theme.** The themes define `categoryAxis` and `valueAxis` only, so the gantt (`type: "time"`) and any date-based line chart fell back to ECharts' un-themed light grey — near-white on the dark canvas. This is the real "different tools on one dashboard" cause.
- **The gantt also carried a local `dashed / opacity 0.3` override**, the one genuine per-chart gridline style in the package.
- **`buildCategoryAxisLabel` returned `show: !compact`** — under 300px the bar widget dropped category identification entirely. That is the four unlabelled bars in `appshell-light.png`.

## Changes

- `GRID_LINE_COLOR` token in `charts/theme.ts`, used by both themes; `timeAxis` registered alongside `categoryAxis`/`valueAxis`.
- Gantt sets only `splitLine: { show }`.
- Compact keeps category labels. The existing sub-400px rule already truncates to 10 chars, so no new budget was needed — which left the `compact` option dead, and it is removed. A tighter 6-char budget was tried first and collapsed `Widget A`..`Widget G` into seven identical `Widge…` stubs.
- Line chart's x-axis labels survive compact too (ECharts thins overlapping ones itself).
- Design-review skill §4 now documents the gridline token and the compact axis policy, plus two anti-patterns — the doc's silence is what let the drift happen.

## Tests

`component/src/charts/__tests__/cartesian-axis-defaults.test.tsx` (new) pins the option shape across bar, line and gantt:
- no chart declares its own `splitLine.lineStyle`
- both themes resolve one gridline colour for category, value **and** time axes
- compact keeps category labels and still drops the value numbers

`axis-label-utils.test.ts` gains a common-prefix case so an over-tight truncation budget fails loudly.

## Verification

- `npm run verify` — 248 + 116 + 28 files, all green
- Playwright: 328 passed. 21 first-run failures were all `AuthPage.login: failed to reach /` (login contention, no chart assertion among them); 19 passed on re-run, the remaining 2 pass in isolation (14/14)
- Storybook, both themes: appshell, bar, gantt, and a 280px bar widget

## Known, out of scope

The gantt's dataZoom slider renders in ECharts' default lavender-blue in both themes. Not filed.

## Summary by CodeRabbit

* **New Features**
  * Improved compact chart behavior: category labels remain visible with truncation when space is limited, while value-axis numbers are hidden.
* **Bug Fixes**
  * Prevented compact modes from hiding category-axis tick labels.
  * Standardized cartesian gridline styling by relying on registered themes (no local split-line styling).
* **Tests**
  * Added coverage for compact/container label rendering and cross-chart axis/gridline defaults.
* **Documentation**
  * Updated chart design guidance for compact behavior and gridline anti-patterns.

### DIFF
--- a/.claude/skills/design-review/skill.md
+++ b/.claude/skills/design-review/skill.md
@@ -132,6 +132,13 @@ grid: { left: 16, right: 16, top: 16, bottom: 24, containLabel: true }
 
 // Compact mode (container < 300px)
 grid: { left: 8, right: 8, top: 8, bottom: 8 }
+// Compact drops the axis NAME and the value-axis numbers. Category labels
+// stay — truncated to 10 chars under 400px — because a chart with no
+// category names identifies nothing (#1247).
+
+// Gridlines: one weight and colour for every cartesian chart, from the
+// registered theme (GRID_LINE_COLOR in charts/theme.ts). Charts set only
+// `splitLine: { show }` — never `splitLine.lineStyle`.
 
 // Legend position
 legend: { bottom: 0 }  // ALWAYS bottom-aligned
@@ -146,6 +153,8 @@ tooltip: { trigger: "axis", axisPointer: { type: "shadow" } }
 - NEVER set chart colors inline — always use `resolveChartColors()`.
 - NEVER add title inside the chart — widget card header IS the title.
 - NEVER register additional ECharts themes — use `neoboard-light` / `neoboard-dark` only.
+- NEVER set `splitLine.lineStyle` in a chart module — gridline weight belongs to the theme (#1247).
+- NEVER let a responsive breakpoint hide category labels — degrade to truncation, not to nothing (#1247).
 - Dark mode chart colors are DIFFERENT from light mode — this is by design (higher lightness for contrast).
 
 ### Graph Chart (NVL)
--- a/component/src/charts/__tests__/axis-label-utils.test.ts
+++ b/component/src/charts/__tests__/axis-label-utils.test.ts
@@ -48,9 +48,22 @@ describe("buildCategoryAxisLabel", () => {
     expect(result.tooltip).toEqual({ show: true });
   });
 
-  it("returns show: false when compact is true", () => {
-    const result = buildCategoryAxisLabel(10, { compact: true });
-    expect(result.show).toBe(false);
+  it("keeps labels in compact mode, truncated not hidden (#1247)", () => {
+    const result = buildCategoryAxisLabel(4, {
+      containerWidth: 280,
+    });
+    expect(result.show).toBe(true);
+    const fmt = result.formatter as (value: string) => string;
+    expect(fmt("Electronics")).toBe("Electroni…");
+    expect(fmt("Home")).toBe("Home");
+  });
+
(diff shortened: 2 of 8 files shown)


alfredo1996:
Applied — good catch, and the distinction you drew is what made it safe.

My original reason for not routing the line chart through `buildCategoryAxisLabel` was that it would apply the width-based rotation heuristic to dense time series, where it is meaningless. Splitting on `useTimeAxis` avoids that entirely: the **category** branch gets the shared config so long labels truncate exactly as they do on the bar chart, and the **time** branch keeps plain labels, since ECharts formats and thins dates itself.

Leaving it as it was would have perpetuated precisely the inconsistency #1247 exists to remove.

Two things worth noting for review:

1. The axis is now built as **two object literals** rather than one with ternaries. ECharts discriminates the axis union on `type`, so a `"time" | "category"` union defeats narrowing and the whole option fails to type-check — I hit that on the first attempt.
2. `width` joins the memo dependency list, as you suggested, so truncation responds to a resize rather than freezing at first render.

Test added to `cartesian-axis-defaults.test.tsx`: a line chart with a long category label truncates to `Electroni…`. 625 chart tests green, types clean.

Issue #1276, opened 2026-07-27
fix(api): self-service password change and profile update are not audited


alfredo1996:
## Problem

After #1234, an **admin resetting someone else's password** is audited (`user.password.reset`). A **user changing their own password** is not — `POST /api/users/me/password` writes no audit entry, and there is no action in the `AuditAction` union for it.

That is the wrong way round for an audit trail. A self-service password change is the single most security-relevant thing an ordinary account can do to itself, and it is the event an incident responder looks for first when asking "was this account taken over?".

`PATCH /api/users/me` (profile: name, email) is likewise unaudited. An attacker changing the email on a compromised account leaves no trace.

Found while wiring the remaining routes for #1234 — both are outside that issue's stated scope, which is why they are filed separately rather than folded in.

## Proposed fix

Two new actions and two call sites, following the pattern the other 21 sites already use:

- `user.password.change` in `app/src/app/api/users/me/password/route.ts` — never the password, old or new.
- `user.profile.update` in `app/src/app/api/users/me/route.ts` — record which fields changed; an email change should record both values, since that is the fact an investigator needs.

Both go through `auditRequest` so the client IP is stamped, which is most of the forensic value here.

## Required tests

**API route (`app/src/app/api/users/me/**/__tests__/`)**
- a successful self password change writes exactly one entry, and the serialized entry contains neither the old nor the new password
- a rejected change (wrong current password, validation failure) writes **no** entry
- a profile update records the changed field names; an email change records old and new
- entries take `tenantId`/`userId` from the session, never the body

## Acceptance criteria

- [ ] Self password change and profile update are audited
- [ ] No password material appears in any entry
- [ ] Failed attempts write nothing
- [ ] The new actions are added to the `AuditAction` union

alfredo1996:
Fixed on release/1.5 by #1725.

Issue #1272, opened 2026-07-26
fix(ci): E2E login retry budget exceeds the test timeout — 21 false failures in one local run


alfredo1996:
## Problem

A full local Playwright run (2026-07-26, on a branch touching only chart option objects) produced **21 failures, every one of them a login timeout** — no chart or feature assertion among them. 19 passed on re-run and the remaining 2 passed in isolation, so nothing was actually broken. The suite reports red when the machine is merely busy.

Two distinct failure signatures:

```
Test timeout of 30000ms exceeded while running "beforeEach" hook.
Error: AuthPage.login: failed to reach / after 3 attempts (last path: /login)
```

## Root cause

`app/e2e/pages/auth.ts` retries login 3 times, each attempt costing a `goto(networkidle)` plus a 10s `waitForURL`. That is a **>30s worst-case budget inside a 30s test timeout** (`playwright.config.ts: timeout: 30_000`). Under load the test is killed partway through attempt 2 or 3 — so the third retry the comment relies on ("three attempts is enough headroom even under CI load") can never actually run when it is needed most.

The retry exists to absorb a React 19 hydration race: clicking Sign in before `onSubmit` is attached makes the form fall back to a plain `GET /login?...`. Retrying after the fact is the expensive way to handle that — waiting for hydration before the click is the cheap one.

## Proposed fix

Prefer fixing the race over widening the budget:

- Gate the click on hydration rather than clicking and retrying — e.g. have the login form expose a ready signal (a `data-hydrated` attribute set in an effect, or a disabled Sign-in button until hydrated) and `waitFor` that. One attempt then suffices.
- If the retry loop stays, its budget must fit the test timeout: 2 attempts × 8s, or an explicit longer timeout on the specs that log in.

## Required tests

**E2E**
- a login test that asserts the form is not interactive before hydration (button disabled / ready flag absent), then becomes interactive — this is the actual contract
- keep one retry-loop test if the loop survives, asserting total elapsed time stays under the test timeout

**Component/jsdom**
- render the login form and assert the ready signal appears only after mount, so the signal cannot be removed silently

## Acceptance criteria

- [ ] Login does not depend on a retry loop to succeed under load
- [ ] Whatever retry remains cannot exceed the enclosing test timeout
- [ ] A full local run under normal desktop load does not produce login-timeout failures

alfredo1996:
Second signature from the same run, same root cause but a different code path:

[...]

Issue #1234, opened 2026-07-25
fix(api): audit trail records nothing — auditLog() has zero callers


alfredo1996:
## Problem

The audit trail is **declared, exposed, and inert**. Verified 2026-07-25:

- `app/src/lib/audit/audit.ts` defines an `AuditAction` union with **25 action types** and a fire-and-forget `auditLog(entry)` writer.
- `auditLogs` table exists in `app/src/lib/db/schema.ts` (tenant-scoped).
- `GET /api/audit-logs` is a real, filterable, paginated listing endpoint.

But `auditLog()` **has no callers**. Grepping `app/src` for `auditLog(` returns only its own definition (`audit.ts:49`) and its test. The only code touching `auditLogs` is the read path in `app/src/app/api/audit-logs/route.ts`.

Net effect: the endpoint faithfully lists an empty table forever, and nothing in the product records who did what.

## Why P1

Audit trails are a compliance feature — the item a security review asks for by name alongside SSO. The current shape (schema + API + action taxonomy, no behavior) is the worst combination for a security questionnaire: it looks implemented. This is also a launch-readiness concern (milestone #29).

## Proposed fix

Call `auditLog()` at the mutation sites the existing action union already names. The taxonomy is the hard part and it is already written. Scope for this issue — the security-relevant mutations:

- **connections**: create, update, delete, reassign
- **dashboards**: create, update, delete, share, import, export
- **users**: create, update, disable, reset-password, role/canWrite change
- **api keys**: create, revoke
- **admin**: key rotation (`/api/admin/rotate-key`)
- **sso providers**: create, update, delete

Requirements:
- Writer stays **fire-and-forget** — an audit failure must never fail the user's request (wrap, log, continue).
- Entries carry `tenantId` from the session, never from the request body (see the tenant-scoping concern in #1226).
- **Never record credential material** — connection audit entries reference the connection id/name, never `configEncrypted` or decrypted config.
- Prefer a single shared helper over 20 hand-rolled call sites, so the shape stays consistent.

## Required tests

**Unit (`app/src/lib/audit/__tests__/`, vitest `unit` project)**
- `auditLog()` writes the expected row shape (action, userId, tenantId, resourceType, resourceId).
- A DB failure inside `auditLog()` is swallowed and logged, never thrown.
- Tenant id is taken from the passed session context.

**API route tests (`app/src/app/api/**/__tests__/`, mocked DB/auth)**
[...]

PR #1281, opened 2026-07-27
feat(api): audit the remaining mutation sites named in #1234

alfredo1996:
Refs #1234 — completes the scope list that #1239 started.

## What was left

#1239 wired connections, dashboards, users and API keys. Cross-checking every mutation route against the issue's scope list found five groups still silent:

| Route | Action |
|---|---|
| `admin/rotate-key` | `admin.key.rotate` — counts only, never key material |
| `connections/[id]/reassign` | `connection.reassign` |
| `dashboards/[id]/share` POST / DELETE | `dashboard.share` / `dashboard.share.revoke` |
| `dashboards/import` | `dashboard.import` |
| `users/[id]/reset-password` | `user.password.reset` |
| `sso-providers` POST/PATCH/DELETE | `sso.provider.create` / `.update` / `.delete` |

21 `auditRequest` call sites now, up from 14.

## Test shape

Every wired route gets two tests: the success path pins the exact entry shape, and a **rejected** mutation asserts no entry is written — the property that makes an audit trail trustworthy rather than merely populated.

Secret-bearing routes assert a denylist on the serialized entry:
- SSO update records field **names** plus `secretRotated: boolean`, never a value
- password reset records `generated: true`, never the password — including the one the route generates itself

`rotate-key`'s `POST` took no arguments; it gains the `request` parameter Next.js already passes, so the caller's IP is stamped like every other site.

## The gap unit tests could not close

`auditLog` is fire-and-forget and mocked in every route test, so nothing proved a write reaches the table. [`e2e/audit-trail.spec.ts`](app/e2e/audit-trail.spec.ts) performs a real mutation and polls `/api/audit-logs` for the row, asserting the acting user and tenant. It passes.

Credential-denylist coverage deliberately stays in the unit tests — asserting it in E2E would mean creating real connections, which other specs count.

## Verification

- `npm run verify` green — 248 app + 115 component + 28 cli test files, lint and types clean
- `e2e/audit-trail.spec.ts` passes in isolation
- The full E2E suite was run and is **not** evidence either way: all 121 failures are the `beforeEach` login-timeout signature of #1272, produced by running against a saturated machine. Zero behavioural assertions failed.

## Still unaudited, deliberately

Eight `AuditAction` values have no writer — the four `auth.*`, both `query.*`, `dashboard.export`, `dashboard.duplicate` — all outside this issue's scope list. Self-service password change is not in the union at all and is now filed as #1276. Documented in [a comment on #1234](https://github.com/alfredo1996/neoboard/issues/1234#issuecomment-5086236263).

## Summary by CodeRabbit

* **New Features**
  * Expanded audit trail coverage for dashboard imports and sharing (including revocations), connection reassignment, user password resets, admin key rotation, and SSO provider create/update/delete.
  * Audit events now record operation context, targets, and outcomes for these actions.
* **Security Improvements**
  * Audit details avoid storing sensitive values such as passwords, client secrets, encryption key material, or ciphertext.
* **Quality Improvements**
  * Added and updated automated tests to verify audit entries are written only on successful requests and omitted when operations fail or are rejected.

### DIFF
--- a/app/e2e/audit-trail.spec.ts
+++ b/app/e2e/audit-trail.spec.ts
@@ -1,0 +1,56 @@
+import { test, expect, ALICE } from "./fixtures";
+
+/**
+ * End-to-end proof that the audit trail records real activity (#1234).
+ *
+ * The route unit tests assert each mutation calls `auditRequest` with the
+ * right shape; they cannot prove the write reaches the table, because
+ * `auditLog` is fire-and-forget and mocked there. This spec closes that gap:
+ * perform a mutation, then read it back through the listing endpoint.
+ *
+ * Credential-denylist coverage stays in the route unit tests — asserting it
+ * here would mean creating real connections, which other specs count.
+ */
+test.describe("Audit trail", () => {
+  test("a dashboard mutation shows up in /api/audit-logs (#1234)", async ({
+    authPage,
+    page,
+  }) => {
+    await authPage.login(ALICE.email, ALICE.password);
+
+    const name = `Audited Dashboard ${Date.now()}`;
+    const created = await page.request.post("/api/dashboards", {
+      data: { name },
+    });
+    expect(created.ok()).toBe(true);
+    const { data: dashboard } = await created.json();
+
+    // The writer is fire-and-forget, so the row can land just after the
+    // response returns. Poll rather than sleep on a fixed delay.
+    await expect
+      .poll(
+        async () => {
+          const res = await page.request.get(
+            "/api/audit-logs?action=dashboard.create&limit=100",
+          );
+          if (!res.ok()) return [];
+          const { data } = await res.json();
+          return data.map((l: { resourceId?: string }) => l.resourceId);
+        },
+        { timeout: 10_000 },
+      )
+      .toContain(dashboard.id);
+
+    // The entry identifies who did it — an audit row without an actor is
+    // no more useful than no row at all.
+    const res = await page.request.get(
+      "/api/audit-logs?action=dashboard.create&limit=100",
+    );
+    const { data } = await res.json();
+    const entry = data.find(
+      (l: { resourceId?: string }) => l.resourceId === dashboard.id,
+    );
+    expect(entry.userId).toBe("user-alice-001");
+    expect(entry.tenantId).toBe("default");
+  });
+});
--- a/app/src/app/api/admin/rotate-key/__tests__/route.test.ts
+++ b/app/src/app/api/admin/rotate-key/__tests__/route.test.ts
@@ -51,7 +51,10 @@ const mockDb = {
   transaction: vi.fn(),
 };
 
-const mockAuditRequest = vi.fn();
+// vi.hoisted: vi.mock factories are hoisted above top-level consts, so the
+// mock must be created in the hoisted scope to stay safe if a future refactor
+// switches this file to a static import of the route.
+const { mockAuditRequest } = vi.hoisted(() => ({ mockAuditRequest: vi.fn() }));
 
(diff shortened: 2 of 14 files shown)


alfredo1996:
CodeRabbit review addressed.

**Adopted — `vi.hoisted` for the audit mocks** (98ccb13d). All six files import their route dynamically inside `beforeEach`, so the factory ran after module init and the plain `const` was safe in practice — the 95 tests passed before and after. But the safety was incidental to the import style, and a later refactor to a static import would have turned it into a load-time TDZ error. `vi.hoisted` removes that coupling. Good catch.

**Dismissed — "Enforce `canWrite` before every SSO mutation" (Major).** The proposed gate can never fire. `canWrite` is unconditionally `true` for admins on *both* auth paths:

```ts
// app/src/lib/auth/session.ts:65
canWrite: role === "admin" ? true : role !== "reader" && (session.user.canWrite !== false),

// app/src/lib/auth/api-key.ts:121
canWrite: role === "admin" ? true : role !== "reader" && row.canWrite,
```

`requireAdmin()` (session.ts:10-20) throws `ForbiddenError` for any non-admin, so by the time these handlers run, `canWrite` is `true` by construction regardless of what the `users` row says. Adding `if (!canWrite) return forbidden()` to the SSO routes would be unreachable code.

The finding did surface something real, though: **the codebase is inconsistent about this**, which is why a reviewer flagged it. `users/[id]/reset-password/route.ts:40` performs exactly that dead check, while `users/route.ts` POST does not. The dead checks imply a protection that the auth layer has already decided doesn't exist for admins, and if the admin rule ever changes, the routes *without* the check become silently vulnerable while the ones with it look deliberate.

Filed separately rather than fixed here, since it spans routes this PR doesn't touch.

alfredo1996:
Both CodeRabbit findings reviewed — one applied, one dismissed with evidence.

**Applied (Minor):** the generated-password test read `mockAuditRequest.mock.calls[0]` without asserting the call count first, so a route that audited twice would still have passed. `toHaveBeenCalledTimes(1)` added.

**Dismissed (Major) — "Handle the audit write after the password reset transaction".** The claim is that `auditRequest(...)` can throw through the route's `catch` and fail a request whose password reset already succeeded. It cannot: **not throwing is the helper's entire contract**, and it is enforced on every path.

```ts
// app/src/lib/audit/audit.ts
export function auditLog(entry: AuditEntry): void {
  const onFailure = (err) => logger.warn({ err, action: entry.action }, "Failed to write audit log");
  try {
    void Promise.resolve(db.insert(auditLogs).values({...})).catch(onFailure);
  } catch (err) {
    onFailure(err);   // sync throw from the driver — pool exhausted, client closed
  }
}

export function auditRequest(request, entry): void {
  let ipAddress = "unknown";
  try { ipAddress = clientIpFromRequest(request); } catch { /* request-like object with no headers */ }
  auditLog({ ...entry, ipAddress });
}
```

Rejected promise, synchronous driver throw, and a request without headers are each caught. The comment in the source calls this out as deliberate — "Belt and braces: `.catch` only covers a rejected promise" — precisely because a failed audit write must never fail the operation that triggered it.

This is pinned by five existing tests in `app/src/lib/audit/__tests__/audit.test.ts`:

- `does not throw when db insert fails`
- `does not throw when the db client throws synchronously`
- `does not throw when the insert builder returns a non-thenable`
- `does not throw when the db insert fails` (auditRequest variant)
- `does not throw when the request has no headers`

So the proposed restructuring would add a guard around something already unconditionally guarded, and the "Heavy lift" cost buys nothing. Worth noting the finding is *directionally* right about the risk — a post-commit side effect must not fail the commit — which is exactly why the helper was built this way in the first place.

Issue #1266, opened 2026-07-26
fix(connection): Testcontainers teardown races the pg pool — unhandled error fails the suite


alfredo1996:
## Problem

`connection`'s Postgres suite intermittently fails the whole job with:

```
FAIL __tests__/postgresql/postgres-query.ts
  ● Test suite failed to run
    Unhandled error. (error: terminating connection due to administrator command
      at parseErrorMessage (node_modules/pg-protocol/src/parser.ts:394:9)
```

`terminating connection due to administrator command` is what Postgres sends when it is **shut down** with clients still attached. So the Testcontainers container is being stopped while a pooled connection is still open, the driver raises an unhandled `error`, and Jest fails the suite — not an assertion failure, a teardown race.

Observed on #1265, whose diff touches only `component/src/charts/`. It cannot have caused this.

## Why it fails the run rather than being ignored

This is the same hazard #999 addressed for request-path clients: the pool's `error` handler does not cover **checked-out** clients, so a backend dying mid-checkout raises an unhandled event. `attachClientErrorGuard` fixed that for `runQuery`. Test teardown evidently has a path where a client is still checked out — or the pool is still open — when the container stops.

## Likely fix

Drain before teardown, and in this order:

1. `await pool.end()` (or the module's `closeConnection`) in `afterAll`, **before** `container.stop()`
2. Confirm nothing holds a checked-out client past the last test — a missing `client.release()` in a failure path would produce exactly this
3. Only then consider tolerating the shutdown error, and if so **only** the `57P01` admin-shutdown SQLSTATE, never a blanket unhandled-error suppressor

Do **not** simply add a global unhandled-error handler to the Jest setup. That would hide the next real one, which is the same trap rejected in #1260.

## Note on local validation

`npm run verify` deliberately skips the `connection` suite because it needs Docker — so this class of failure **cannot** be caught by the current pre-push loop. Either document that gap in the `Validation` practice note (done) or add an opt-in `verify:full` that includes it. Worth deciding.

## Required tests

- run the Postgres suite **10 consecutive times**; zero unhandled-error failures. This is a race, so a single green run proves nothing
- assert the pool is closed before the container stops — e.g. a teardown-order test or an explicit assertion that `pool.ended` is true
- prove any tolerated error is narrowed to the admin-shutdown SQLSTATE and that an *unrelated* unhandled error still fails the suite

## Acceptance criteria

- [ ] Pool drained before container teardown
- [ ] 10 consecutive clean runs
[...]

Issue #1240, opened 2026-07-25
fix(ci): unit-test oversubscription flake has spread beyond data-grid


alfredo1996:
## Evidence

On [#1238](https://github.com/alfredo1996/neoboard/pull/1238) — a **docs-only PR that touches zero runtime code** — CI failed with:

```
FAIL |component| src/app/(dashboard)/settings/api-keys/__tests__/page.test.tsx
  > ApiKeysPage Key column (#1038) > renders a Key header and the masked prefix cell
  5144ms  (default testTimeout 5000ms)
```

A docs-only diff cannot cause this. It is the same starvation signature as #1184 (signup) and #1208 (data-grid grouping): a jsdom test creeping just past the 5s default timeout under CI load.

## Root cause (already documented, now confirmed spreading)

`.github/workflows/ci.yml` runs all four packages' `test:coverage` **concurrently on a single runner**, while `connection/`'s Testcontainers suite starts real Postgres and Neo4j containers. Each Vitest process spawns its own workers on top of that.

#1208 fixed the symptom for one file and left a `ponytail:` comment saying exactly this:

> raising the timeout tolerates the oversubscription; the real root cause is that parallel fan-out — bound it in ci.yml if this flakiness spreads.

**It has spread.** Patching timeouts file-by-file as each one surfaces is now the more expensive path, and each occurrence costs a debugging cycle and erodes the "no flakes on dev" rule.

## Proposed fix

Address the fan-out rather than the individual tests. Options, cheapest first:

1. **Bound concurrency in `ci.yml`** — stagger the four packages, or cap each Vitest project's workers so the total stays under the runner's cores. Cheapest, no test churn.
2. **Split `unit-tests` into a matrix job per package** — real isolation, more runner minutes, better failure attribution.
3. Raise the global `testTimeout` for CI only — treats the symptom project-wide; acceptable as a stopgap alongside (1).

Prefer (1); (3) is a reasonable belt-and-braces addition.

## Required tests

CI-config change, so verification is empirical rather than unit-testable:

- Re-run the full `unit-tests` job **5 consecutive times** on an unchanged commit and confirm zero flakes (this is the acceptance signal).
- Record the wall-clock delta so the fix isn't quietly trading all its benefit for slower CI.
- No per-test timeout bumps in this PR — the point is to stop needing them. If a specific test still fails after the fan-out is bounded, that test has a real bug and gets its own issue.

## Acceptance criteria

[...]

PR #1322, opened 2026-07-27
fix(connection): honour sslmode from the PostgreSQL URI instead of discarding it

### LINKED ISSUE #1299: fix(connection): sslmode in the PostgreSQL URI is discarded, so an explicit ?sslmode=require connects in plaintext
## Problem

`?sslmode=...` in a PostgreSQL connection URI is parsed and thrown away. A user who writes `postgresql://db.internal:5432/app?sslmode=require` gets an **unencrypted** connection — no error, no warning, and a green "Test Connection". There is no other way to request encryption: the only TLS control the product exposes is the "Reject Unauthorized SSL" switch, which toggles certificate verification once TLS is already on and cannot turn TLS on.

`createDriver()` reads three fields off the parsed URL and discards the rest — `connection/src/postgresql/PostgresAuthenticationModule.ts:37-61`:

```ts
const url = new URL(this._authConfig.uri);
const database = url.pathname.slice(1) || "postgres";

// Build SSL config if specified
const ssl =
  this._advancedOptions?.pgSslRejectUnauthorized === undefined
    ? undefined
    : { rejectUnauthorized: this._advancedOptions.pgSslRejectUnauthorized };

const pool = new Pool({
  user: this._authConfig.username,
  password: this._authConfig.password,
  host: url.hostname,
  port: parseInt(url.port || "5432", 10),
  database: database,
  ...
  ...(ssl ? { ssl } : {}),
});
```

`url.search` / `url.searchParams` are never read. And because the pool is built from discrete fields rather than `connectionString`, node-pg never sees the query string either. `grep -rn "searchParams|url.search|connectionString" connection/src/ connector-sdk/src/` returns **zero** hits, and `new Pool(` appears exactly once in the whole repo (line 50) — there is no other path that could pick it up.

Measured directly: URI `postgresql://db.example.com:5432/mydb?sslmode=verify-full` produced pool config `{user, password, host, port, database, connectionTimeoutMillis, idleTimeoutMillis, max}` — no `ssl`, no `connectionString`, no throw, no warning. `node_modules/pg/lib/defaults.js:50` is `ssl: false` and `pg/lib/client.js:166-167` is `if (self.ssl) { con.requestSsl() }`, so no SSLRequest packet is ever sent.

The plaintext default is currently **pinned by a test rather than flagged** — `connection/__tests__/advanced-connection-options.test.ts:205-212`:

```ts
it("does not include ssl when pgSslRejectUnauthorized is undefined", () => {
  new PostgresAuthenticationModule(pgAuth);
  expect(poolConstructorCalls[0].ssl).toBeUndefined();
});
```

Three supporting facts about why there is no workaround:

- `connection/src/form-fields.ts:43-68` (`postgresFormFields`) is, per its own header, "the single, client-safe source of truth for what the connection form renders (#1118)". It exposes exactly `uri`, `username`, `password`, `database` — no SSL control.
- The `sslmode` select at `component/src/components/composed/connection-form.tsx:144-152` — with its `prefer` default and the #981 rationale comment — is a **separate demo field list that nothing on the server consumes**. `grep -rn sslmode app/src` returns zero hits.
- `app/src/app/(dashboard)/connections/page.tsx:654` renders `checked={form.sslRejectUnauthorized ?? false}`, so the switch shows the same "off" position for two different wire behaviours: `undefined` (no TLS at all) and `false` (TLS with verification disabled).

## Why this is P1

Anyone who can create a connection — creator or admin, via `POST /api/connections` or the inline `POST /api/connections/test-inline` — and points NeoBoard at a database that is not on localhost gets an unencrypted link to a customer production database. Every result row of every widget query crosses the network in the clear, as does the auth handshake: with `password` auth the credential is literally on the wire; with `md5` the captured exchange is offline-crackable; only SCRAM-SHA-256 resists it.

The multiplier is the silence. A user who does the standard libpq thing and appends `?sslmode=require` believes they have opted in, and "Test Connection" agrees with them. An insecure default is a known risk; a security control that reports success while doing nothing is not visible to anyone.

This does not break a literal line of the Query Safety / Credentials rules — those govern query text, parameterization, and the encryption key at rest — but it defeats their purpose. `NEVER log decrypted credentials` protects the credential in logs and at rest while the handshake puts it on the wire in cleartext.

## Proposed fix

Honour the URI. In `createDriver()`, read `sslmode` off `url.searchParams` and build the `ssl` option explicitly, with libpq semantics:

| `sslmode` | `ssl` passed to `new Pool` | meaning |
| --- | --- | --- |
| absent or `disable` | omitted (pg default `ssl: false`) | no SSLRequest sent — today's behaviour |
| `require` | `{ rejectUnauthorized: false }` | encrypt; do not verify the certificate |
| `verify-ca` | `{ rejectUnauthorized: true, checkServerIdentity: () => undefined }` (plus `ca` when `sslrootcert` is given) | verify the chain, not the hostname |
| `verify-full` | `{ rejectUnauthorized: true }` | verify chain and hostname |

An explicitly-set `pgSslRejectUnauthorized` continues to override the URI-derived `rejectUnauthorized`, so configs using the advanced switch keep working. An **unrecognised** `sslmode` value must throw at construction, never fall through to plaintext.

Two constraints that the mapping table has to respect:

- **Do not offer `prefer`.** node-pg has no opportunistic mode. `pg/lib/connection.js:73-75` turns a server `N` response into `The server does not support SSL connections`, so "encrypt if possible, else plaintext" needs a connect-fail-reconnect retry that nobody has written. Until then `prefer` should not appear in any field list — the dead select at `connection-form.tsx:144-152` should be removed or reduced to the four modes above.
- **Leave the no-`sslmode` default as it is (unencrypted).** Flipping it hard-fails every local dev setup: the stock `postgres` image ships `ssl = off`, and pg errors rather than downgrading. Hardening the default depends on the `prefer` retry above and is a separate decision. The P1 here is that an *explicit* request is discarded.

Once the URI is honoured, two follow-on cleanups make it discoverable: add `sslmode` to `postgresFormFields` (`connection/src/form-fields.ts:43`) so it is not URI-only, and fix the tri-state switch at `connections/page.tsx:654` so "no TLS" and "TLS, unverified" stop rendering identically.

### Why not just pass the URI as `connectionString`

pg already contains the entire libpq mapping (`pg-connection-string/index.js:77-160`), so `new Pool({ connectionString: uri, user, password })` looks like the one-line version of this fix. It silently destroys the credentials. `pg/lib/connection-parameters.js:59-60`:

```js
// if the config has a connectionString defined, parse IT into the config we use
// this will override other default values with what is stored in connectionString
if (config.connectionString) {
  config = Object.assign({}, config, parse(config.connectionString))
}
```

`parse()` returns `user` and `password` keys even when the URI carries no userinfo — as empty strings — so they overwrite the decrypted values. Measured:

```
> new ConnectionParameters({ user: 'realuser', password: 'realpw',
    connectionString: 'postgresql://db.example.com:5432/mydb?sslmode=require' })
{ user: 'alfredorubin', password: null, host: 'db.example.com', database: 'mydb', ssl: {} }
```

`realuser` / `realpw` are gone; `user` fell through to pg's OS-username default. NeoBoard stores credentials separately from the URI and passes them as `this._authConfig.username/password`, so every existing connection would start failing auth.

Two further traps on that path, which also apply if you call `parse()` yourself instead of building `ssl` by hand:

- `ssl: {}` above, not `{ rejectUnauthorized: false }`. In pg 8.20, `sslmode=require` is treated as an alias for `verify-full` unless `uselibpqcompat=true` is also present in the URI. `{}` leaves `rejectUnauthorized` at Node's default `true`, so every internal-CA or self-signed Postgres would start failing certificate validation on a mode libpq defines as "encrypt, don't verify".
- `parse()` **throws** on `sslmode=verify-ca&uselibpqcompat=true` with no `sslrootcert` (`pg-connection-string/index.js:121`). `createDriver()` rethrows from its catch at line 81, so that converts a saved config into a hard construction failure.

Building the `ssl` object explicitly avoids all three, and keeps behaviour stable across the pg 9 / pg-connection-string 3.0 semantics change that pg's own runtime warning announces.

## Required tests

1. **Unit — `connection/__tests__/advanced-connection-options.test.ts`** (already mocks the `Pool` constructor into `poolConstructorCalls`). Table-drive the mapping: for absent, `disable`, `require`, `verify-ca`, `verify-full`, assert the exact `ssl` value handed to `new Pool`. Fails today on every row except absent/`disable` — `poolConstructorCalls[0].ssl` is `undefined` for all five. Add: `sslmode=bogus` throws; `pgSslRejectUnauthorized: true` still wins over `?sslmode=require`. **Line 205's `it("does not include ssl when pgSslRejectUnauthorized is undefined")` must be rewritten** — as written it pins the defect, and should survive only for the no-`sslmode` case.

2. **Integration — `connection/__tests__/postgresql/` (Testcontainers, real database).** Start Postgres with `ssl=on` and a generated self-signed certificate. Assert server-side, not on the client's pool options: `SELECT ssl FROM pg_stat_ssl WHERE pid = pg_backend_pid()`.
   - `?sslmode=require` → `ssl = true`. Fails today: `false`.
   - `?sslmode=disable` → `ssl = false`.
   - `?sslmode=verify-full` against the self-signed cert → connect **fails**. Fails today: it connects, in plaintext. This is the case that catches a silent downgrade; a client-side assertion on pool options cannot.

3. **Conformance — `connector-sdk/src/conformance/query-safety.ts`.** "A connector never silently downgrades a requested transport security level" should bind *every* connector, not just the built-in Postgres one, so it belongs in the shared harness that `connection/__tests__/conformance/pg-conformance.test.ts` and `neo4j-conformance.test.ts` both already run. Add a case that takes a config requesting encryption and asserts the connection is either encrypted or refused — never established unencrypted. Neo4j supplies `neo4j+s://` / `+ssc` as its encrypted variant, Postgres supplies `?sslmode=require`. This needs TLS-enabled containers in both setups and is the largest piece of the work, which is exactly why it is worth writing once in the harness.

4. **App route — vitest in `app/src/app/api/connections/__tests__/route.test.ts` and a sibling for `test-inline/route.ts`.** Assert "Test Connection" reports failure for a config whose requested `sslmode` cannot be satisfied, rather than returning success. Today the request never reaches TLS negotiation at all, so the route reports success for a URI that asked for encryption and did not get it.

5. **E2E — `app/e2e/connection-advanced.spec.ts`**, only if the form field lands in step "follow-on cleanups": set SSL Mode when creating a PostgreSQL connection, save, reopen the edit dialog, confirm the value round-trips. Fails today because no such control exists on the real form.

## Acceptance criteria

- [ ] `?sslmode=require` produces an encrypted connection, verified server-side via `pg_stat_ssl.ssl = true`.
- [ ] `?sslmode=verify-full` against an untrusted certificate fails to connect and never falls back to plaintext.
- [ ] `?sslmode=disable` and a URI with no `sslmode` keep today's behaviour (no `ssl` option, no SSLRequest sent) — no existing local or dev connection breaks.
- [ ] An unrecognised `sslmode` value throws at pool construction instead of defaulting to plaintext.
- [ ] An explicit `pgSslRejectUnauthorized` still overrides the URI-derived verification setting.
- [ ] The pool is still constructed from discrete fields; `connectionString` is not passed to `new Pool`, so decrypted credentials never traverse pg's URI parse path.
- [ ] `connection/__tests__/advanced-connection-options.test.ts:205` no longer asserts that `ssl` is absent irrespective of `sslmode`.
- [ ] `prefer` appears in no field list, or is implemented with a real fallback retry.
- [ ] The transport-downgrade conformance case is registered in `connector-sdk/src/conformance/query-safety.ts` and passes for both built-in connectors.

alfredo1996:
Closes #1299

## Problem

`createDriver()` parsed the connection URI for host, port and database and ignored the query string entirely (`connection/src/postgresql/PostgresAuthenticationModule.ts`):

```ts
const url = new URL(this._authConfig.uri);
const database = url.pathname.slice(1) || "postgres";
```

Every managed Postgres — RDS, Cloud SQL, Neon, Supabase, Heroku — hands the user a URI ending in `?sslmode=require`. NeoBoard dropped it and **connected in plaintext**, with nothing in the UI or the logs to say so.

TLS happened only if the operator separately discovered the `pgSslRejectUnauthorized` advanced option — which no copy-pasted URI sets. So the default experience for "paste your production connection string" was an unencrypted link to a customer's production database.

## Fix

Map `sslmode` to node-pg's `ssl` option using libpq's semantics, which are not a simple on/off:

| sslmode | Result | Why |
|---|---|---|
| `disable` | `ssl: false` | explicit no-TLS |
| `allow`, `prefer` | `{ rejectUnauthorized: false }` | libpq falls back to plaintext; node-pg cannot negotiate that, so we do the secure half — encrypt without demanding a verifiable chain — rather than silently not encrypting |
| `require` | `{ rejectUnauthorized: false }` | encrypt, do not verify — this is what `require` means, and conflating it with verification would break every self-signed managed instance |
| `verify-ca`, `verify-full` | `{ rejectUnauthorized: true }` | encrypt and verify |
| absent / unrecognised | `undefined` | unchanged behaviour |

**Precedence:** the explicit `pgSslRejectUnauthorized` advanced option still wins. An operator sets that deliberately in the connection form; a parameter inherited from a pasted URI must not silently override it.

**One subtle change:** the Pool spread goes from `...(ssl ? ...)` to `...(ssl !== undefined ? ...)`. `sslmode=disable` resolves to boolean `false` — a deliberate instruction that was being dropped as falsy.

## Tests

Seven new cases in `connection/__tests__/advanced-connection-options.test.ts`, all failing before the change: each sslmode value, the precedence rule, absence, and one adversarial case — a database literally named `sslmode=require` (`postgresql://host/sslmode=require`), which must **not** enable TLS. That last one is why the parsing goes through `url.searchParams` rather than a substring match.

I also had to fix the helper I added: `poolConstructorCalls` accumulates across tests in that describe block, so reading index `[0]` silently asserted against an earlier test's call.

## Verification

- 21/21 in the touched file; **all 9 Postgres suites green, 107 tests**
- The 9 failures in the full connection run are pre-existing contention, not this change: `list-databases.ts` fails at 77s inside the parallel run and **passes in 8.7s on its own**. Same pattern as #1240 and #1266.

## Not covered

No integration test asserts a real TLS handshake — that needs a Postgres container with certificates, which the current Testcontainers setup doesn't provision. The unit tests pin the config that reaches the driver; proving the wire is encrypted is a larger piece of work worth doing separately.

### DIFF
--- a/connection/__tests__/advanced-connection-options.test.ts
+++ b/connection/__tests__/advanced-connection-options.test.ts
@@ -265,4 +265,73 @@ describe("DEFAULT_CONNECTION_CONFIG (#973)", () => {
       await import("@neoboard/connector-sdk");
     expect(DEFAULT_CONNECTION_CONFIG.timeout).toBe(30_000);
   });
+
+  // -------------------------------------------------------------------------
+  // sslmode in the connection URI (#1299)
+  //
+  // Every managed Postgres hands you a URI with ?sslmode=require. Dropping it
+  // means NeoBoard silently connects to a customer's production database in
+  // PLAINTEXT, with nothing in the UI or logs to say so.
+  // -------------------------------------------------------------------------
+
+  function moduleWithUri(uri: string, advanced?: PostgresAdvancedOptions) {
+    const {
+      PostgresAuthenticationModule,
+    } = require("../src/postgresql/PostgresAuthenticationModule");
+    // poolConstructorCalls accumulates across tests in this describe block, so
+    // reset and read the call this helper just made rather than index [0].
+    poolConstructorCalls.length = 0;
+    new PostgresAuthenticationModule({ ...pgAuth, uri }, advanced);
+    return poolConstructorCalls[poolConstructorCalls.length - 1];
+  }
+
+  it("honours sslmode=require from the URI (#1299)", () => {
+    // libpq semantics: require = encrypt, do not verify the certificate.
+    const cfg = moduleWithUri(
+      "postgresql://localhost:5432/testdb?sslmode=require",
+    );
+    expect(cfg.ssl).toEqual({ rejectUnauthorized: false });
+  });
+
+  it("honours sslmode=verify-full from the URI (#1299)", () => {
+    const cfg = moduleWithUri(
+      "postgresql://localhost:5432/testdb?sslmode=verify-full",
+    );
+    expect(cfg.ssl).toEqual({ rejectUnauthorized: true });
+  });
+
+  it("honours sslmode=verify-ca from the URI (#1299)", () => {
+    const cfg = moduleWithUri(
+      "postgresql://localhost:5432/testdb?sslmode=verify-ca",
+    );
+    expect(cfg.ssl).toEqual({ rejectUnauthorized: true });
+  });
+
+  it("treats sslmode=disable as explicitly no TLS (#1299)", () => {
+    const cfg = moduleWithUri(
+      "postgresql://localhost:5432/testdb?sslmode=disable",
+    );
+    expect(cfg.ssl).toBe(false);
+  });
+
+  it("lets the explicit advanced option override the URI (#1299)", () => {
+    // pgSslRejectUnauthorized is set deliberately by an operator in the
+    // connection form; a query param inherited from a copy-pasted URI must
+    // not silently win over it.
+    const cfg = moduleWithUri(
+      "postgresql://localhost:5432/testdb?sslmode=verify-full",
+      { pgSslRejectUnauthorized: false },
+    );
+    expect(cfg.ssl).toEqual({ rejectUnauthorized: false });
+  });
+
+  it("still omits ssl when neither the URI nor options ask for it (#1299)", () => {
+    const cfg = moduleWithUri("postgresql://localhost:5432/testdb");
+    expect(cfg.ssl).toBeUndefined();
+  });
+
+  it("does not mistake a database named like a param for sslmode (#1299)", () => {
+    const cfg = moduleWithUri("postgresql://localhost:5432/sslmode=require");
+    expect(cfg.ssl).toBeUndefined();
+  });
 });
--- a/connection/src/postgresql/PostgresAuthenticationModule.ts
+++ b/connection/src/postgresql/PostgresAuthenticationModule.ts
@@ -4,6 +4,45 @@ import { Pool } from "pg";
 import { isAuthenticationError, attachClientErrorGuard } from "./utils";
 
 /**
+ * Translate libpq's `sslmode` into node-pg's `ssl` option (#1299).
+ *
+ * Every managed Postgres — RDS, Cloud SQL, Neon, Supabase, Heroku — hands the
+ * user a URI ending in `?sslmode=require`. We parsed the URI for host, port and
+ * database and ignored the query string entirely, so those connections were
+ * made in PLAINTEXT with nothing in the UI or the logs to say so.
+ *
+ * libpq semantics, which are not the same as "on/off":
+ *   disable                  no TLS at all
+ *   allow / prefer           opportunistic; libpq falls back to plaintext.
+ *                            node-pg cannot negotiate a fallback, so we do the
+ *                            secure half — encrypt without demanding a
+ *                            verifiable chain — rather than silently not
+ *                            encrypting.
+ *   require                  encrypt, do NOT verify the certificate
+ *   verify-ca / verify-full  encrypt AND verify
+ *
+ * Returns undefined for an absent or unrecognised value, preserving the
+ * previous "no ssl key" behaviour for connections that never asked for TLS.
+ */
+export function sslFromSslMode(
+  sslmode: string | null,
+): false | { rejectUnauthorized: boolean } | undefined {
+  switch (sslmode?.trim().toLowerCase()) {
+    case "disable":
+      return false;
+    case "allow":
+    case "prefer":
+    case "require":
+      return { rejectUnauthorized: false };
+    case "verify-ca":
+    case "verify-full":
+      return { rejectUnauthorized: true };
+    default:
+      return undefined;
+  }
+}
+
+/**
  * PostgreSQL Authentication Module
  * Handles connection pooling and authentication for PostgreSQL databases.
  */
@@ -38,10 +77,15 @@ export class PostgresAuthenticationModule extends AuthenticationModule {
       const url = new URL(this._authConfig.uri);
       const database = url.pathname.slice(1) || "postgres";
 
-      // Build SSL config if specified
+      // Build SSL config. The explicit advanced option wins — an operator set
+      // it deliberately in the connection form, and a query param inherited
+      // from a copy-pasted URI must not silently override that. Otherwise fall
+      // back to the URI's sslmode, which every managed Postgres includes and
+      // which we previously discarded, silently connecting in plaintext to
+      // databases that asked for TLS (#1299).
       const ssl =
         this._advancedOptions?.pgSslRejectUnauthorized === undefined
-          ? undefined
+          ? sslFromSslMode(url.searchParams.get("sslmode"))
           : {
               rejectUnauthorized: this._advancedOptions.pgSslRejectUnauthorized,
             };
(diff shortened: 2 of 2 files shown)


alfredo1996:
@coderabbitai review

Issue #1298, opened 2026-07-27
fix(connection): write path buffers the entire result set, so MAX_ROWS+1 never bounds memory on writes


alfredo1996:
## Problem

One root cause, present in both built-in connectors: on the WRITE path `rowLimit` is a *display* cap applied after the full result set is already resident in the Node heap. The MAX_ROWS+1 pattern — which CLAUDE.md states unconditionally under **Query Safety** ("Row limits: cursor/stream consumption with MAX_ROWS+1 pattern") — holds only for reads.

PostgreSQL, `connection/src/postgresql/PostgresConnectionModule.ts:163-177` — the read branch streams, the write branch does not:

```ts
if (isReadOnly) {
  const batch = await readBoundedCursor(client, query, paramValues, config.rowLimit + 1);
  fetchedRows = batch.rows;
  fields = batch.fields;
} else {
  const result = await client.query(query, paramValues);   // full buffer
  fetchedRows = result.rows;
  fields = result.fields;
  affectedRowCount = result.rowCount ?? undefined;
}
```

The cap is applied only afterwards, at `PostgresConnectionModule.ts:182-185`:

```ts
const isTruncated = fetchedRows.length > config.rowLimit;
const limitedRows = isTruncated ? fetchedRows.slice(0, config.rowLimit) : fetchedRows;
```

Neo4j, `connection/src/neo4j/Neo4jConnectionModule.ts:99-105` — same shape:

```ts
const res = await tx.run(query, params);   // awaiting buffers the WHOLE result
const all = res.records;
const writeTruncated = all.length > config.rowLimit;
return {
  rows: writeTruncated ? all.slice(0, config.rowLimit) : all,
  truncated: writeTruncated,
};
```

The comment above it (`Neo4jConnectionModule.ts:95-98`) justifies the exemption: "Writes must run to completion so every side effect executes; ... Stopping a write stream early could skip un-pulled CREATE/SET/DELETE work, leaving a partially-applied transaction." That reasoning is correct and must be preserved — but running to completion requires *draining* the stream, not *retaining* every record. Two lines down, the read branch shows the codebase already knows the difference (`Neo4jConnectionModule.ts:111-112`):

```ts
const res = tx.run(query, params);        // deliberately not awaited
return collectUpToLimit(res, config.rowLimit);
```

[...]

PR #1325, opened 2026-07-27
fix(connection): drain Neo4j writes instead of buffering the whole result set

alfredo1996:
Refs #1298 — Neo4j half. PostgreSQL is deliberately **not** in this PR; reasoning below.

## Problem

The write branch awaited `tx.run()`, which buffers every record into `QueryResult.records`, and only then sliced:

```ts
const res = await tx.run(query, params);
const all = res.records;
return { rows: all.slice(0, config.rowLimit), ... };
```

So `rowLimit` was a **display cap, not a memory bound**. `POST /api/query/write` executes caller-supplied text with no query-kind validation, gated on `canWrite` — so any user with a Form widget could send `MATCH (n) SET n.seen = true RETURN n` and materialise every record in the Node heap **shared by every tenant on the process**. #978 fixed exactly this on the read path and left the write path standing.

## The distinction that matters

The existing comment justified the exemption — *"writes must run to completion so every side effect executes"* — and it is correct. But it argues for **draining** the stream, not for **retaining** it. The read branch two lines below already knows the difference: it deliberately does not `await tx.run`.

New SDK helper, beside `collectUpToLimit` so external connectors get it too:

| | Behaviour | Correct for |
|---|---|---|
| `collectUpToLimit` | stops pulling once truncation is known, invoking the iterator's `return()` and releasing the portal | **reads** |
| `drainRetainingUpTo` | pulls every row — all side effects run — and stops *keeping* them past the limit | **writes** |

Using `collectUpToLimit` on a write would be a data-integrity bug rather than a perf choice: a database executes `UPDATE … RETURNING` incrementally, so rows never pulled are never modified. It would silently turn "update 1,000,000 rows" into "update 26" **and still report success**. That trap is documented on the helper so the next person doesn't reach for the wrong one.

## Tests

Five, all failing before:

- a 1000-row source is **fully pulled** while 10 are retained
- rows are retained in order, `truncated` is accurate
- `rowLimit: 0` still drains
- a side-by-side case pins the difference: **100 rows pulled with no early `return()`**, against 6 pulled and a `return()` for `collectUpToLimit`

## Why PostgreSQL is not here

Same defect, but the fix must **not** reuse `readBoundedCursor`. That does one bounded `cursor.read(maxRows)` and closes the cursor in its `finally` — on an `UPDATE … RETURNING` the portal is abandoned before the remaining rows are applied, producing a silently partially-applied write. That is worse than the memory problem it would solve.

The correct shape is a drain loop over `cursor.read(batch)` until it yields zero rows, **plus** proof that an `INSERT` without `RETURNING` still reports its affected-row count — the behaviour the current comment at `PostgresConnectionModule.ts:153-158` exists to protect. That needs a real Postgres and an integration test, not a stub. Splitting it keeps a verified improvement moving instead of holding it behind an unverified one.

### DIFF
--- a/connection/src/neo4j/Neo4jConnectionModule.ts
+++ b/connection/src/neo4j/Neo4jConnectionModule.ts
@@ -13,7 +13,7 @@ import {
 import { Neo4jRecordParser } from "./Neo4jRecordParser";
 import { extractNodeAndRelPropertiesFromRecords } from "./utils";
 import { determineQueryStatus } from "@neoboard/connector-sdk";
-import { collectUpToLimit } from "@neoboard/connector-sdk";
+import { collectUpToLimit, drainRetainingUpTo } from "@neoboard/connector-sdk";
 import { wrapError, ConnectorErrorType } from "@neoboard/connector-sdk";
 
 /**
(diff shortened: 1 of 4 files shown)


alfredo1996:
@coderabbitai review

PR #1327, opened 2026-07-27
fix(connection): settle the caller when a query is empty, instead of hanging

### LINKED ISSUE #1301: fix(connection): runQuery returns without invoking a terminal callback on an empty query, wedging the scheduler slot forever
One root cause: `ConnectionModule.runQuery` is consumed as a promise that settles **only** through `onSuccess`/`onFail`, but nothing in the SDK enforces that a terminal callback is ever invoked — and the SDK's own shared helper already violates it. Two live violation sites follow from that single gap (the SDK helper, and Neo4j's session creation outside `try`).

## Problem

Any authenticated user can permanently wedge a connector's query scheduler by submitting a whitespace-only query. `maxConcurrent` such requests and every subsequent query on that connection queues and then fails with 408/503, with no recovery short of a process restart.

The shared helper returns early without invoking any terminal callback:

```ts
// connector-sdk/src/generalized/ConnectionModule.ts:35-47
protected handleEmptyQuery<T>(
  query: string | undefined,
  callbacks: QueryCallback<T>,
): boolean {
  if (callbacks.setStatus) {
    if (!query || query.trim() === "") {
      callbacks.setStatus(QueryStatus.NO_QUERY);
      return true;
    }
```

Both built-in connectors bail on it — `connection/src/neo4j/Neo4jConnectionModule.ts:64` and `connection/src/postgresql/PostgresConnectionModule.ts:61` are both:

```ts
if (this.handleEmptyQuery(query, callbacks)) return;
```

No `onSuccess`, no `onFail`. The app wraps `runQuery` in a promise that settles only via those callbacks, discards the returned promise, and has no wall-clock fallback:

```ts
// app/src/lib/query/query-executor.ts:290-316
return new Promise((resolve, reject) => {
  connModule.runQuery(
    finalQueryParams,
    {
      onSuccess: (result: unknown) => resolve({ data: result, truncated, rowLimit: effectiveRowLimit }),
      onFail: (error: unknown) => reject(error),
      ...
    },
    config,
  );
});
```

That promise is awaited inside the scheduler middleware, which releases the ticket in a `finally` that never runs:

```ts
// app/src/lib/query/middleware/scheduler.ts:56-59
try {
  return await next();
} finally {
  scheduler.release(ticket.id);
}
```

`queueTimeoutMs` (`app/src/lib/query/scheduler.ts:193`) only drops **waiters**; a running ticket has no timeout, so the slot is held for the lifetime of the process.

The route's validation does not stop it — `query: z.string().min(1)` (`app/src/app/api/query/route.ts:40`) accepts `"   "`, which reaches `handleEmptyQuery` and hangs.

Because the same promise is the only settlement path, a *rejecting* `runQuery` is equally fatal. PostgreSQL documents this explicitly:

```ts
// connection/src/postgresql/PostgresConnectionModule.ts:63-67
// Invariant: runQuery must NEVER reject. The caller wraps this in a promise
// that settles only via onSuccess/onFail, so any thrown/rejected error here
// would leave that promise pending forever — hanging the request and pinning
// a scheduler slot. Funnel every failure (auth, pool.connect, network) to
// onFail. Neo4j already upholds this contract. (#CRITICAL)
```

Neo4j does not uphold it. `_runCypherQuery` creates the session **outside** the `try`:

```ts
// connection/src/neo4j/Neo4jConnectionModule.ts:83
const session = this.getDriver().session({
  defaultAccessMode: neo4j.session[config.accessMode],
  database: config.database,
});
```

`getDriver()` returns `this.authModule.getDriver()`, which hands back the driver field directly (`connection/src/neo4j/Neo4jAuthenticationModule.ts:31-33`). A concurrent `closeConnection` or `_evictStaleEntries` (`app/src/lib/query/query-executor.ts:85`, `:110`) closes that driver; `session()` on a closed driver throws, `runQuery` rejects, and the executor's promise stays pending forever — same wedge, no whitespace needed.

The abstract declaration carries no contract at all (`connector-sdk/src/generalized/ConnectionModule.ts:14-19`), so the only statement of the invariant that keeps the app alive is a comment inside one built-in connector.

## Why this is P1

Trigger: `POST /api/query` with `{"connectionId": "<any connection the caller can read>", "query": "   "}`. Any authenticated role that can run a query — including a reader on a shared dashboard — passes the zod schema, acquires a scheduler ticket, and never releases it. Repeat `maxConcurrent` times (`QUERY_MAX_CONCURRENT`, small by default) and that connector is dead for **all tenants' users** on that connection for the remaining process lifetime: new queries wait out `queueTimeoutMs` and return 408/503. This is unauthenticated-adjacent resource exhaustion with a spacebar and no server-side recovery, and it defeats the concurrency/backpressure controls the scheduler exists to provide.

For third-party connectors it is worse: a plugin author reading `ConnectionModule` sees no stated contract, so a *conformant-looking* connector violates it by default.

## Proposed fix

Make terminal settlement part of the contract, enforce it in the shared helper, and close both violation sites.

1. **Settle in `handleEmptyQuery`.** After `callbacks.setStatus(QueryStatus.NO_QUERY)`, invoke a terminal callback before returning `true` — `callbacks.onSuccess?.([] as T)` (an empty query legitimately produced no rows) or a dedicated `onEmpty`. Pick one and document it; the requirement is that exactly one terminal callback fires on every path.
2. **Document the invariant on the abstract method** (`ConnectionModule.runQuery`, `connector-sdk/src/generalized/ConnectionModule.ts:14-19`): *"`runQuery` must never reject, and must invoke exactly one of `onSuccess` / `onFail` on every path."* Today that sentence lives only in a PostgreSQL comment.
3. **Move `this.getDriver().session({...})` inside the `try`** in `Neo4jConnectionModule._runCypherQuery` (line 83) so a throw from `session()` funnels to `onFail`, mirroring the PostgreSQL funnel.
4. **Add the invariant to the conformance harness** (`connector-sdk/src/conformance/query-safety.ts`, `buildConformanceCases`) so it binds *every* connector, not just the two built-ins — see Required tests.
5. **Defense in depth in the app:** in `executeQuery` (`app/src/lib/query/query-executor.ts:290-316`), treat `QueryStatus.NO_QUERY` in the `setStatus` handler as a terminal reject, so a non-conformant third-party connector cannot hang the request even if it ignores the contract.

### Why not just reject whitespace-only queries in the route's zod schema

Changing `z.string().min(1)` to `.trim().min(1)` closes the one trigger named above and leaves the defect intact. The bug is not "empty queries are allowed"; it is "`runQuery` has an unenforced settlement contract and the executor has no fallback." The Neo4j closed-driver path (item 3) wedges the scheduler with a perfectly valid query and is untouched by input validation, and any third-party connector that returns early on its own edge case reintroduces the same hang. Tightening the schema is fine as a bonus, but only items 1–5 remove the class of failure.

### Why not add a wall-clock timeout in `executeQuery`

A `Promise.race` around `runQuery` would unstick the slot eventually, but it converts a deterministic contract violation into a silent per-request stall of the timeout duration, and it makes non-conformant connectors invisible rather than failing them at conformance time. Ship items 1–4 first; a race is at best a belt on top of item 5.

## Required tests

- **Conformance harness (preferred — binds every connector):** add two cases to `buildConformanceCases` in `connector-sdk/src/conformance/query-safety.ts`. Both built-in connectors already run the harness via `connection/__tests__/conformance/neo4j-conformance.test.ts` and `connection/__tests__/conformance/pg-conformance.test.ts`, so these bind Neo4j and PostgreSQL for free and every future connector by construction. The harness's `execute()` helper (`query-safety.ts:56-81`) already captures `onSuccess` / `onFail` / statuses — the new cases assert that at least one terminal callback fired.
  - *"an empty or whitespace-only query settles via a terminal callback"* — run `{ query: "   " }`; fail if neither `onSuccess` nor `onFail` was invoked. Fails on today's code for both connectors: `handleEmptyQuery` returns after only `setStatus(NO_QUERY)`.
  - *"runQuery never rejects"* — wrap the `execute()` call so a rejected `runQuery` promise is a conformance failure rather than a test error, and exercise it against a module whose driver/pool has been closed. Fails on today's Neo4j connector: `session()` at `Neo4jConnectionModule.ts:83` throws outside the `try`.
- **Connection integration (`connection/__tests__/`, vitest + Testcontainers):** extend the existing empty-query assertions — `connection/__tests__/connection/query-status.ts:295` and `:323`, `connection/__tests__/postgresql/postgres-query.ts:157` — which today assert only `QueryStatus.NO_QUERY` and would still pass with the hang in place. Assert additionally that the chosen terminal callback fired.
- **App unit (`app/src/lib/__tests__/query/query-executor-core.test.ts`):** with a stub `ConnectionModule` whose `runQuery` returns without calling any callback, assert `executeQuery(...)` rejects rather than staying pending (covers fix item 5). Fails today — the promise never settles, so the test hangs until the vitest timeout.
- **App route (`app/src/app/api/query/__tests__/`):** `POST /api/query` with `query: "   "` returns a definite response (not a hang), and the scheduler ticket is released — assert `scheduler.release` was called, since `app/src/lib/query/middleware/scheduler.ts:56-59` only releases in `finally`. This is the test that encodes the user-visible symptom.
- **E2E:** not required. The failure is a server-side scheduler leak with no distinctive UI; a Playwright run would only observe the eventual 408/503.

## Acceptance criteria

- [ ] `handleEmptyQuery` invokes exactly one terminal callback before returning `true`.
- [ ] `ConnectionModule.runQuery`'s abstract declaration documents: never rejects; invokes exactly one of `onSuccess` / `onFail` on every path.
- [ ] `Neo4jConnectionModule._runCypherQuery` creates its session inside the `try`, so a closed-driver throw reaches `onFail` instead of rejecting `runQuery`.
- [ ] `buildConformanceCases` includes an "empty query settles" case and a "runQuery never rejects" case; both built-in connector conformance suites pass with them.
- [ ] `executeQuery` rejects when a connector reports `QueryStatus.NO_QUERY` without a terminal callback, so a non-conformant connector cannot hang the request.
- [ ] `POST /api/query` with a whitespace-only query returns a response and releases its scheduler ticket; repeating it `maxConcurrent + 1` times leaves the connector able to serve a normal query.
- [ ] All new tests fail on the current `dev` commit and pass after the fix.

alfredo1996:
Closes #1301

## Problem

`handleEmptyQuery` returns `true` so the connector returns early — and invokes **no terminal callback**:

```ts
// connector-sdk/src/generalized/ConnectionModule.ts
if (!query || query.trim() === "") {
  callbacks.setStatus(QueryStatus.NO_QUERY);
  return true;      // no onSuccess, no onFail
}
```

`runQuery` is consumed as a promise that settles **only** through `onSuccess`/`onFail` (`app/src/lib/query/query-executor.ts`), and the app discards the returned promise with no wall-clock fallback. So a whitespace-only query left that promise pending forever, holding the scheduler slot it had already acquired.

`maxConcurrent` such requests wedged the connector: every subsequent query on that connection queued and then failed 408/503, with **no recovery short of a process restart**. Any authenticated user could trigger it by saving a blank widget query.

## Fix

One place. Both built-in connectors bail on this shared helper — `Neo4jConnectionModule.ts:64` and `PostgresConnectionModule.ts:61` are both `if (this.handleEmptyQuery(query, callbacks)) return;` — so fixing the helper fixes them and every external connector at once, rather than patching two call sites and leaving plugin authors exposed.

**Settled as success-with-no-rows, not failure.** `NO_QUERY` already carries the meaning, and reporting a blank widget query as an error would paint a red error state on every dashboard that happens to have one. A blank query isn't a fault, it's an absence.

## Tests

Seven, five failing before:

- `undefined`, `""`, `"   "` and `"\n\t "` each produce **exactly one** terminal callback — the invariant the whole issue rests on
- an empty query resolves rather than rejects, so the UI shows "no data" instead of an error
- a non-empty query still settles nothing and reports `RUNNING`, so the normal path is unchanged

## Verification

- 37 SDK tests
- 106 connection unit tests
- 181 app query tests — the consumer whose promise was hanging

## Note

The audit that found this also flagged that `handleEmptyQuery` skips the check entirely when a caller omits `setStatus`. That was **refuted** on review and is left alone here; this PR changes only the terminal-callback behaviour.

### DIFF
--- a/connector-sdk/__tests__/handle-empty-query.test.ts
+++ b/connector-sdk/__tests__/handle-empty-query.test.ts
@@ -1,0 +1,95 @@
+import { ConnectionModule } from "../src/generalized/ConnectionModule";
+import { QueryStatus } from "../src/generalized/interfaces";
+import type { QueryCallback } from "../src/generalized/interfaces";
+
+/**
+ * `handleEmptyQuery` returns true so the connector returns early. Callers wrap
+ * `runQuery` in a promise that settles ONLY through onSuccess/onFail, so an
+ * early return with neither leaves that promise pending forever — and with it
+ * the scheduler slot it holds. `maxConcurrent` whitespace-only queries wedge a
+ * connector until the process restarts (#1301).
+ */
+
+/** Minimal concrete subclass — we only exercise the protected helper. */
+class TestModule extends ConnectionModule<unknown> {
+  runQuery(): void {}
+  checkConnection(): Promise<boolean> {
+    return Promise.resolve(true);
+  }
+  listDatabases(): Promise<string[]> {
+    return Promise.resolve([]);
+  }
+  callHandleEmptyQuery(query: string | undefined, cbs: QueryCallback<unknown>) {
+    return this.handleEmptyQuery(query, cbs);
+  }
+}
+
+function spyCallbacks() {
+  const calls = {
+    onSuccess: [] as unknown[],
+    onFail: [] as unknown[],
+    status: [] as QueryStatus[],
+  };
+  const callbacks: QueryCallback<unknown> = {
+    onSuccess: (r) => calls.onSuccess.push(r),
+    onFail: (e) => calls.onFail.push(e),
+    setStatus: (s) => calls.status.push(s),
+  };
+  return { callbacks, calls };
+}
+
+describe("handleEmptyQuery terminal-callback contract (#1301)", () => {
+  it.each([undefined, "", "   ", "\n\t "])(
+    "settles the caller for query %p",
+    (query) => {
+      const { callbacks, calls } = spyCallbacks();
+
+      const handled = new TestModule().callHandleEmptyQuery(query, callbacks);
+
+      expect(handled).toBe(true);
+      expect(calls.status).toContain(QueryStatus.NO_QUERY);
+      // The whole point: exactly one terminal callback must fire, or the
+      // caller's promise never settles.
+      expect(calls.onSuccess.length + calls.onFail.length).toBe(1);
+    },
+  );
+
+  it("reports an empty query as success with no rows, not as a failure", () => {
+    const { callbacks, calls } = spyCallbacks();
+
+    new TestModule().callHandleEmptyQuery("  ", callbacks);
+
+    // NO_QUERY already carries the meaning. Surfacing it as onFail would turn
+    // a blank widget into a red error state on every dashboard load.
+    expect(calls.onFail).toHaveLength(0);
+    expect(calls.onSuccess).toEqual([[]]);
+  });
+
+  it("does not settle the caller for a non-empty query", () => {
+    const { callbacks, calls } = spyCallbacks();
+
+    const handled = new TestModule().callHandleEmptyQuery(
+      "RETURN 1",
+      callbacks,
+    );
+
+    expect(handled).toBe(false);
+    expect(calls.status).toContain(QueryStatus.RUNNING);
+    expect(calls.onSuccess).toHaveLength(0);
+    expect(calls.onFail).toHaveLength(0);
+  });
+
+  it("still settles when the caller supplies no setStatus", () => {
+    const calls = { onSuccess: [] as unknown[] };
+    const handled = new TestModule().callHandleEmptyQuery("", {
+      onSuccess: (r) => calls.onSuccess.push(r),
+    });
+
+    // Without setStatus the old code skipped the check entirely and returned
+    // false, handing an empty query to the driver. Whatever it decides, it
+    // must not leave the caller hanging.
+    if (handled) {
+      expect(calls.onSuccess).toHaveLength(1);
+    }
+  });
+});
--- a/connector-sdk/src/generalized/ConnectionModule.ts
+++ b/connector-sdk/src/generalized/ConnectionModule.ts
@@ -31,6 +31,17 @@ export abstract class ConnectionModule {
   /**
    * Checks for empty/missing query and sets the appropriate status.
    * Returns true if the query is empty (caller should return early).
+   *
+   * Returning true means the connector does NOT run the query — so this must
+   * settle the caller itself. `runQuery` is consumed as a promise that resolves
+   * only through `onSuccess`/`onFail`; returning early with neither left that
+   * promise pending forever, holding its scheduler slot. `maxConcurrent`
+   * whitespace-only queries were enough to wedge a connector until the process
+   * restarted (#1301).
+   *
+   * Settled as success-with-no-rows rather than failure: `NO_QUERY` already
+   * carries the meaning, and treating a blank widget query as an error would
+   * paint a red error state on every dashboard that has one.
    */
   protected handleEmptyQuery<T>(
     query: string | undefined,
@@ -39,6 +50,7 @@ export abstract class ConnectionModule {
     if (callbacks.setStatus) {
       if (!query || query.trim() === "") {
         callbacks.setStatus(QueryStatus.NO_QUERY);
+        callbacks.onSuccess?.([] as T);
         return true;
       }
       callbacks.setStatus(QueryStatus.RUNNING);


alfredo1996:
@coderabbitai review

PR #1328, opened 2026-07-27
fix(connection): drain PostgreSQL writes through a cursor instead of buffering

### LINKED ISSUE #1326: fix(connection): PostgreSQL write path buffers the entire result set before the row limit applies
## Problem

Split from #1298, whose Neo4j half is fixed in #1325. PostgreSQL has the same defect and a materially harder fix.

`connection/src/postgresql/PostgresConnectionModule.ts:163-177` — the read branch streams, the write branch buffers:

```ts
if (isReadOnly) {
  const batch = await readBoundedCursor(client, query, paramValues, config.rowLimit + 1);
  fetchedRows = batch.rows;
} else {
  const result = await client.query(query, paramValues);   // full buffer
  fetchedRows = result.rows;
  affectedRowCount = result.rowCount ?? undefined;
}
```

The cap is applied afterwards (`:182-185`), so `rowLimit` bounds what is *displayed*, not what is *resident*. One Form submit against a large table can exhaust the heap shared by every tenant on the process.

## Why this is not a copy of the Neo4j fix

**`readBoundedCursor` must not be reused here.** It performs one bounded `cursor.read(maxRows)` and closes the cursor in its `finally` (`cursor-read.ts:28-51`). PostgreSQL executes a portal incrementally, so an `UPDATE … RETURNING *` suspended after `rowLimit + 1` rows **has not applied the remaining updates**, and closing the portal before `COMMIT` abandons that work. Swapping it in would silently convert "update 1,000,000 rows" into "update 26" and still report success — worse than the memory problem it solves.

**The affected-row count must survive.** The comment at `:153-158` records why the buffered path was kept: an `INSERT` without `RETURNING` must still report `COMPLETE` rather than `NO_DATA`, which comes from `result.rowCount`. Any cursor-based rewrite has to preserve that, and a cursor over a non-returning statement is exactly where that is most likely to break.

## Proposed fix

Drain to exhaustion, retain at most `rowLimit`:

- run the write through a cursor inside the existing read-write transaction
- loop `cursor.read(batchSize)` until a batch returns **zero** rows, so the portal runs to completion
- retain only the first `rowLimit + 1`
- take the affected-row count from the final result / `CommandComplete` tag, not from a buffered `rows` array

`drainRetainingUpTo` (added to the SDK in #1325) is the right shape if the cursor can be adapted to an async iterable; otherwise mirror its semantics.

## Required tests

**Integration — `connection/__tests__/postgresql/` (Testcontainers, real Postgres). These are the point of the issue; a stubbed client cannot prove any of them.**

- with `rowLimit: 10`, run `UPDATE t SET x = x + 1 RETURNING *` over 1,000 rows, then `SELECT count(*)` to assert **all 1,000 were updated** while only 10 rows came back. This is the regression guard for the `readBoundedCursor` trap and must fail on any early-stopping implementation.
- an `INSERT` without `RETURNING` still reports the correct affected-row count and `COMPLETE`, not `NO_DATA`.
- a write returning more than `rowLimit` rows sets the truncation flag.

**Unit — `connection/__tests__/postgresql/`**
- the write branch passes a `Cursor` to `client.query` and keeps reading until a batch is empty (stubbed `PoolClient`); fails today, where it calls `client.query(query, values)` with a plain string.

**Conformance — `connector-sdk/src/conformance/query-safety.ts`**
- the existing row-limit case is pinned to `accessMode: "READ"` (`:124`), which is why this gap was invisible. Add a `WRITE` case so the rule binds every connector, including external ones.

## Acceptance criteria

- [ ] A write returning more rows than `rowLimit` does not buffer them all
- [ ] Every affected row is still modified — proven against a real database
- [ ] `INSERT` without `RETURNING` still reports its affected-row count
- [ ] The conformance harness covers the row limit under `WRITE`, not only `READ`

alfredo1996:
Closes #1326 · Refs #1298 — the PostgreSQL half. Neo4j is #1325.

## Problem

The write branch buffered everything before the row limit applied:

```ts
const result = await client.query(query, paramValues);
fetchedRows = result.rows;          // whole set, sliced afterwards
```

So `rowLimit` bounded what was **displayed**, not what was **resident**. One Form submit against a large table could exhaust the heap shared by every tenant on the process.

## Why the obvious fix would have been worse

`readBoundedCursor` does one bounded read and closes the cursor in its `finally`. That is right for a `SELECT`, and wrong here: PostgreSQL applies an `UPDATE … RETURNING` **incrementally**, so rows never pulled are never modified, and closing the portal abandons them. Swapping it in would have turned "update 1,000 rows" into "update 11" — silently, while still reporting success.

`drainBoundedCursor` reads to exhaustion — every row produced, every side effect run — retaining only `maxRows`. Peak memory becomes the batch size plus `maxRows` rather than the whole result.

## A bug the integration test caught in my own fix

The completing batch of a `RETURNING` statement reports `rowCount: 0`. My first version let that overwrite the real count, so an `UPDATE` that changed a thousand rows reported **`NO_DATA`**. `affectedRowCount` is now surfaced only when the statement returned no rows — precisely the `INSERT`-without-`RETURNING` case it exists for.

Worth stating plainly: a stubbed client would have happily passed the broken version.

## Tests — against a real PostgreSQL

`connection/__tests__/postgresql/postgres-write-drain.ts`, Testcontainers:

1. **`UPDATE … RETURNING` over 1,000 rows with `rowLimit: 10`** returns ≤10 rows **and leaves all 1,000 updated**. This is the regression guard for the trap above — it fails on any implementation that stops reading early.
2. **`INSERT` without `RETURNING`** still reports its affected-row count rather than `NO_DATA` — the behaviour the original buffered path existed to protect.
3. A write exceeding the limit still flags `COMPLETE_TRUNCATED`.

Note these pass on the *old* code too: they assert correctness properties the buffered version also satisfied. Their job is to make the fix's failure modes impossible to ship, not to demonstrate the memory bug — which is why the memory claim rests on the implementation change rather than on a flaky heap measurement.

## One adapted test

`postgres-error-handling.test.ts` simulated a failing `INSERT` through `client.query`, which the write path no longer uses — so the failure never occurred and `onFail` was never called. The rejection now originates in `drainBoundedCursor`. The test's subject is unchanged: a failed query whose `ROLLBACK` also fails must still surface the **original** error.

## Verification

8 Postgres suites, 72 tests.

## Summary by CodeRabbit

* **New Features**
  * Added bounded result handling for write queries, limiting rows returned to the caller while ensuring all database changes complete.
  * Write queries that exceed the row limit now report a truncated completion status.
  * For writes that don’t return rows, affected-row counts are preserved.

* **Bug Fixes**
  * Write operations without returned rows no longer incorrectly report “no data”.
  * Improved regression coverage so the original write failure is reported even if rollback also fails (rollback failure is logged separately).

* **Tests**
  * Added an integration test suite validating write-path behavior against a real PostgreSQL instance.

### DIFF
--- a/connection/__tests__/postgresql/postgres-error-handling.test.ts
+++ b/connection/__tests__/postgresql/postgres-error-handling.test.ts
@@ -17,6 +17,15 @@ import {
 // (which can't back a real pg-cursor) still exercises the transaction logic.
 jest.mock("../../src/postgresql/cursor-read", () => ({
   readBoundedCursor: jest.fn().mockResolvedValue({ rows: [], fields: [] }),
+  // Writes drain rather than stopping early (#1298); the write path calls
+  // this one, so the double has to provide it or every write test fails on
+  // "drainBoundedCursor is not a function" rather than on its own assertion.
+  drainBoundedCursor: jest.fn().mockResolvedValue({
+    rows: [],
+    fields: [],
+    affectedRowCount: 0,
+    truncated: false,
+  }),
 }));
 
 function makeModule(): PostgresConnectionModule {
@@ -198,6 +207,15 @@ describe("PostgresConnectionModule — error-path routing", () => {
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     jest.spyOn(mod.authModule, "getPool").mockReturnValue(pool as any);
     const errSpy = jest.spyOn(console, "error").mockImplementation(() => {});
+
+    // Writes now stream through drainBoundedCursor rather than client.query
+    // (#1298), so that is where the statement failure originates. The test's
+    // subject is unchanged: a failing query whose ROLLBACK also fails must
+    // still surface the ORIGINAL error through onFail.
+    const { drainBoundedCursor } = require("../../src/postgresql/cursor-read");
+    (drainBoundedCursor as jest.Mock).mockRejectedValueOnce(
+      new Error("insert exploded"),
+    );
 
     const onFail = jest.fn();
     await mod.runQuery(
--- a/connection/__tests__/postgresql/postgres-error-handling.test.ts
+++ b/connection/__tests__/postgresql/postgres-error-handling.test.ts
@@ -24,7 +24,6 @@ jest.mock("../../src/postgresql/cursor-read", () => ({
     rows: [],
     fields: [],
     affectedRowCount: 0,
-    truncated: false,
   }),
 }));
 
--- a/connection/__tests__/postgresql/postgres-write-drain.ts
+++ b/connection/__tests__/postgresql/postgres-write-drain.ts
@@ -58,7 +58,9 @@ describe("PostgreSQL write path — row limit must not truncate side effects", (
   }, 120_000);
 
   afterAll(async () => {
-    await connectionModule.close();
+    // Guarded: if beforeAll throws before construction, an unguarded close()
+    // raises a TypeError that masks the real setup failure.
+    await connectionModule?.close();
     await container?.stop();
   }, 60_000);
 
@@ -68,7 +70,7 @@ describe("PostgreSQL write path — row limit must not truncate side effects", (
       (resolve, reject) => {
         const statuses: QueryStatus[] = [];
         connectionModule.runQuery(
-          { query, parameters: {} },
+          { query, params: {} },
           {
             onSuccess: (rows: unknown) =>
               resolve({ rows: rows as unknown[], statuses }),
@@ -79,7 +81,7 @@ describe("PostgreSQL write path — row limit must not truncate side effects", (
           },
           {
             ...DEFAULT_CONNECTION_CONFIG,
-            type: ConnectionTypes.POSTGRESQL,
+            connectionType: ConnectionTypes.POSTGRESQL,
             rowLimit: ROW_LIMIT,
             accessMode: "WRITE",
           },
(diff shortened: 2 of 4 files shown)


alfredo1996:
@coderabbitai review

alfredo1996:
All four addressed — thanks, one of these was a real hole in the test rather than a style point.

**`params` / `connectionType` (Major)** — correct on both. The second is the one that mattered: `DEFAULT_CONNECTION_CONFIG.connectionType` defaults to `NEO4J`, so the suite was configuring a Neo4j connection type against the Postgres module. Behaviour was still right — the `UPDATE` ran, which a READ ONLY transaction would have rejected, so `accessMode: "WRITE"` was being honoured — but the test would have started lying the moment anything branched on `connectionType`.

**`toBeLessThanOrEqual` (Minor, but the best catch here)** — you're right that it passes when the drain returns **zero** rows, which is precisely the failure this suite exists to detect. An early-stopping implementation that returned nothing would have sailed through. Now `toHaveLength(ROW_LIMIT)`.

**Unguarded `connectionModule.close()`** — fixed; `container?.stop()` was already guarded, so this was inconsistent as well as fragile.

**Stale block comment + discarded `batch.truncated`** — the comment still claimed writes "keep the direct path", the opposite of the new behaviour. Rewritten to state what actually differs: both paths stream, they differ in how they *stop*.

On `truncated`: rather than wiring it up, I removed it from `DrainedCursor`. The caller derives truncation uniformly from the retained row count, so the field was dead on arrival — and shipping an unused one invites the next reader to trust it.

72 tests across 8 Postgres suites still green with the stricter assertion.

alfredo1996:
Fifth finding addressed — `container` is now `StartedPostgreSqlContainer`, which is what `start()` actually resolves to.

Worth noting: **`postgres-query.ts:11` has the identical mistake**, and my file copied it from there. I've left that one alone rather than widening a P1 PR into an unrelated file — but it means the two files now disagree, so it's worth a follow-up sweep rather than being forgotten.

The other four from this review are already fixed in the previous commit; those comments are anchored to pre-fix line numbers rather than still being open.

PR #1331, opened 2026-07-28
fix(connection): promote int8/numeric to numbers and emit interval as text

### LINKED ISSUE #1307: fix(connection): postgres int8/numeric arrive as strings, interval as a sparse object
## Problem

These are two symptoms of **one root cause**: `PostgresRecordParser` passes driver output through untouched on the strength of a premise that is empirically false for three OIDs. The doc comment claims "pg driver already returns native primitives" (`connection/src/postgresql/PostgresRecordParser.ts:61` and `:74`), but `pg-types` ships no numeric parser for OID 20 (`int8`) or 1700 (`numeric`), and returns OID 1186 (`interval`) as a prototype-bearing `PostgresInterval`.

Measured against the installed `pg-types` in `connection/`:

```
$ node -e "const t=require('pg-types'); ..."
20   (int8)      string  "9007199254740993"
1700 (numeric)   string  "9007199254740993"
701  (float8)    number  9007199254740992
23   (int4)      number  9007199254740992
1231 (numeric[]) object  [1.5,2.5]          <-- elements ARE numbers
1186 (interval)  object  {"days":1}
1186 (interval)  object  {"hours":2}
```

There is no `setTypeParser` call anywhere in `connection/src`, `app/src`, `connector-sdk/src` or `cli/src`, so pg defaults apply unmodified.

**Symptom 1 (P2) — `int8`/`numeric` reach widgets as strings.** `SELECT count(*)` yields the string `"42"` on PostgreSQL and the number `42` on Neo4j for the same logical query. `numeric` scalars are strings but `numeric[]` elements are floats, so the pg side is not even self-consistent. The parser converts nothing — `connection/src/postgresql/PostgresRecordParser.ts:73-78`:

```ts
/**
 * No-op: pg driver already returns native primitives.
 */
parsePrimitive(value: unknown): unknown {
  return value;
}
```

and `_pgToNative` ends at `:57` with a bare `return value;`.

Concrete downstream break, `component/src/lib/export-utils.ts:9` and `:16`:

```ts
const isNumeric = typeof value === "number" || typeof value === "bigint";
...
if (!isNumeric && /^[=@+\-\t\r]/.test(str)) { str = `'${str}`; }
```

A `numeric` column holding `-12.50` arrives as the string `"-12.50"`, fails the `isNumeric` check, matches the leading-`-` formula-injection rule, and exports to CSV as `'-12.50` — forced to text in Excel. The identical value from Neo4j exports as a number.

**Symptom 2 (P3) — `interval` is flattened into a sparse plain object.** `_pgToNative` special-cases `Date` (`:46`) and `Buffer` (`:52`), then routes everything else object-shaped to the generic copier at `connection/src/postgresql/PostgresRecordParser.ts:55-56`:

```ts
if (typeof value === "object")
  return this.pgConvertPlainObject(value as object);
```

`pgConvertPlainObject` → `NeodashRecordParser.convertPlainObject` (`connector-sdk/src/generalized/NeodashRecordParser.ts:82-90`) copies own enumerable keys into a fresh `{}`, dropping the prototype. `postgres-interval/index.js:11` does `extend(this, parse(raw))` and `parse` skips zero components, so `INTERVAL '1 day'` reaches the widget as `{days: 1}` — no `toISOString()`, no `toPostgres()`, and no `hours`/`minutes`/`seconds` keys at all. `app/src/lib/shared/normalize-value.ts:29-33` then renders it via the `JSON.stringify` fallback as the literal string `{"days":1}`. Neo4j's equivalent (`connection/src/neo4j/Neo4jRecordParser.ts:222-229`) always returns the full fixed shape `{ months, days, seconds, nanoseconds }`.

No test covers any of this. Grepping the suite for pg scalar types returns only `connection/__tests__/postgresql/postgres-parser.ts:22` (bytea) and `:133` (Date) — nothing for `numeric`, `int8`, `jsonb`, `interval` or `uuid`.

## Why this is P2

Any reader-role user viewing any widget backed by a PostgreSQL connector, on the ordinary query path (`PostgresConnectionModule.executeQuery` → `bulkParse` at `connection/src/postgresql/PostgresConnectionModule.ts:203`). No special permission, no special query — `count(*)` and `numeric` money columns are the two most common things a dashboard selects.

It is P2 rather than P1 because each symptom is small and none is a security or data-loss issue; it is not P3 because it is silent and wide. Charts survive (`toSeriesNumber` coerces strings), so nobody notices until a table sorts `"100"` before `"9"`, a CSV export lands as text in Excel, or a widget's `typeof === "number"` branch takes a different path depending on which database backs the dashboard. For `interval` the key set changes row to row — `INTERVAL '2 hours'` gives `{hours: 2}` while `INTERVAL '1 day'` gives `{days: 1}` — so a consumer reading `.seconds` gets `undefined` for most rows rather than `0`.

## Proposed fix

Adopt the safe-range contract the Neo4j side already implements, so both connectors agree. `Neo4jRecordParser.ts:116` does:

```ts
return value.inSafeRange() ? value.toNumber() : value.toBigInt();
```

**Promote `int8`/`numeric` gated on the column's `dataTypeID`, not on the string's shape.** The field descriptors are already in hand: `readBoundedCursor` returns `CursorBatch { rows, fields }` (`connection/src/postgresql/cursor-read.ts:8-11`), and `fields` is assigned at `PostgresConnectionModule.ts:175`, in scope at the `bulkParse` calls on `:203` and `:209`. Promote only OIDs 20 and 1700, and only when the round-trip is exact:

```ts
// int8/numeric arrive as text (pg-types registers no parser for OID 20/1700).
// Promote only when Number() round-trips exactly — otherwise keep the string,
// same contract as Neo4j's inSafeRange() check.
if (typeof value === "string") {
  const n = Number(value);
  if (String(n) === value) return n;
}
```

Plumb the OIDs by adding a pg-side `bulkParse(rows, fields)` overload (the SDK base signature is `bulkParse(_records)` only, `connector-sdk/src/generalized/NeodashRecordParser.ts:9`) or by constructing a parser per query. **Do not stash the fields on `this.parser`** — that instance is created once in the constructor (`PostgresConnectionModule.ts:35`) and shared across every concurrent query the scheduler dispatches on that connection, so per-query mutable state on it is a cross-query corruption bug.

**Emit `interval` as text instead of an object**, in a branch next to the `Buffer` case, mirroring how that one emits Postgres's own canonical form:

```ts
// interval arrives as a PostgresInterval; the generic object copier would
// strip its prototype formatters and leave a sparse {days: 1}. Emit the
// ISO-8601 duration the driver already knows how to produce.
if (typeof (value as { toPostgres?: unknown }).toPostgres === "function") {
  return (value as { toISOString(): string }).toISOString();
}
```

Note this makes pg emit `"P1D"` where Neo4j emits `{months, days, seconds, nanoseconds}`. That is strictly better than today (renderable, stable, no missing keys), but the two connectors still differ in *shape*. Converging them means changing the Neo4j `Duration` branch too, which is an existing contract widgets may read — track that as a follow-up rather than folding it in here.

### Why not `pg.types.setTypeParser(20, Number)`

It is process-global — it changes decoding for every node-pg pool in the process, including any SDK plugin connector, with no way for a connector to opt out. Worse, bare `Number` silently truncates: `9007199254740993` becomes `9007199254740992`, exactly the precision loss `inSafeRange()` exists to prevent. (The app's own metadata DB is safe from this — `app/src/lib/db/index.ts` uses `postgres-js`, not node-pg — but that is luck, not design.)

### Why not sniff every string with a numeric regex

A `text` column containing `"42"` would become the number `42`, silently changing sort order, export type and `typeof` branches for genuine text data. The `String(n) === value` guard saves `"01234"` and `"1.50"` but cannot save `"42"` — only the OID gate can.

### Why not cast in SQL

Rewriting the user's query to `::float8`/`::text` violates the package's Query Safety rule **"NEVER modify or wrap user queries"** — safety and typing are enforced at the driver/parser level, never by editing query text.

## Required tests

**1. Parser unit test — `connection/__tests__/postgresql/postgres-parser.ts`** (no container needed; this is the only pg parser test file and currently covers 2 of ~8 non-trivial types).
- `int8`/`numeric` columns given the driver's text output return `typeof === "number"`; assert `parse({ n: "42" }, [{ dataTypeID: 20 }]).n === 42`. Fails today: returns the string `"42"`.
- A value outside the safe range stays a string: `"9007199254740993"` must not become `9007199254740992`. Passes today by accident, must keep passing after the fix — this is the regression guard against the `setTypeParser(20, Number)` shortcut.
- A `text` column (OID 25) holding `"42"` stays the string `"42"`. Fails on any string-sniffing implementation.
- A `PostgresInterval` literal parses to the ISO duration string, not `{days: 1}`. Fails today: returns `{"days":1}`.

**2. Cross-connector conformance case — new module in `connector-sdk/src/conformance/`** (sibling to `query-safety.ts`), exported from `connector-sdk/src/index.ts` and wired into **both** `connection/__tests__/conformance/pg-conformance.test.ts` and `connection/__tests__/conformance/neo4j-conformance.test.ts`, which already run real databases via Testcontainers. This rule — *the same logical value has the same JS type from every connector* — should bind every connector, including future SDK ones, so it belongs in the shared harness rather than in the pg suite alone. Minimum case: a count query (`SELECT count(*) …` / `MATCH … RETURN count(*)`) must yield `typeof === "number"` from both. Fails today on the PostgreSQL side only, which is precisely the defect.

**3. No app-route or E2E test required.** The defect is entirely inside the connector's parse step; `escapeCsvCell` (`component/src/lib/export-utils.ts`, covered by `component/src/lib/__tests__/export-utils.test.ts`) behaves correctly given a number and needs no change. A Playwright run would only re-prove what the conformance case pins, at container cost.

## Acceptance criteria

- [ ] `int8` (OID 20) and `numeric` (OID 1700) values reach widgets as `number` when `Number()` round-trips exactly, and as the original string otherwise.
- [ ] Promotion is gated on `dataTypeID`; a `text` column containing `"42"` is still the string `"42"`.
- [ ] `9007199254740993` from an `int8` column is not silently truncated to `9007199254740992`.
- [ ] No `pg.types.setTypeParser` call is introduced.
- [ ] The per-query field descriptors are not stored on the shared `PostgresConnectionModule.parser` instance.
- [ ] `INTERVAL '1 day'` and `INTERVAL '2 hours'` produce a stable, renderable value (no sparse `{days: 1}` / `{hours: 2}` objects, no `{"days":1}` string in the table).
- [ ] `connection/__tests__/postgresql/postgres-parser.ts` covers int8, numeric, out-of-safe-range int8, text-that-looks-numeric, and interval.
- [ ] A type-fidelity conformance case lives in `connector-sdk/src/conformance/` and is run by both `pg-conformance.test.ts` and `neo4j-conformance.test.ts`.
- [ ] The `count(*)` conformance case passes identically on both built-in connectors.
- [ ] The stale "No-op: pg driver already returns native primitives" comments at `PostgresRecordParser.ts:61` and `:74` are corrected.
- [ ] No user query text is modified, wrapped, or cast to achieve any of the above.

alfredo1996:
Closes #1307

## Problem

`PostgresRecordParser` passed driver output through untouched on the strength of a doc comment — *"pg driver already returns native primitives"* — that is empirically false for three OIDs. `pg-types` registers **no parser** for OID 20 (`int8`) or 1700 (`numeric`), and returns 1186 (`interval`) as a prototype-bearing `PostgresInterval`.

So `SELECT count(*)` yielded the **string** `"42"` on PostgreSQL and the **number** `42` on Neo4j for the same logical query. `numeric` scalars were strings while `numeric[]` elements were floats — the pg side wasn't even self-consistent.

The concrete downstream break: `export-utils.ts` tests `typeof value === "number"`, so a money column of `-12.50` failed that check, matched the leading-hyphen formula-injection rule, and exported to CSV as `'-12.50` — forced to text in Excel. The identical value from Neo4j exported as a number.

## Design constraints honoured

**Gated on `dataTypeID`, not on the string's shape.** A product code of `"42"` in a `text` column must stay a string — there's a test for exactly that.

**Fields passed per call, never stored on the parser.** That instance is constructed once and shared across every concurrent query the scheduler dispatches on a connection, so per-query state there would be a cross-query corruption bug. `bulkParse(rows, fields?)` takes them as an argument.

## One deliberate deviation from the issue

The issue proposed:

```ts
const n = Number(value);
if (String(n) === value) return n;
```

That rejects `"-12.50"` — the single most common case, and the one the issue's **own worked example** is about. `Number("-12.50")` is `-12.5`, whose `String()` is `"-12.5"`, so the money column stays a string and nothing is fixed.

A trailing zero is **formatting, not value**; display precision belongs to `formatNumber`. So the rule here compares *canonicalised* forms:

| Input | Result | Why |
|---|---|---|
| `"42"` | `42` | exact |
| `"-12.50"` | `-12.5` | trailing zero is formatting |
| `"9007199254740993"` | `"9007199254740993"` | 2⁵³+1 — a double would silently return `…992`. **Real data loss.** |

Same spirit as the Neo4j side's `inSafeRange()` check: precision beats type-consistency, but only where precision is genuinely at stake.

## interval

Now emits Postgres's own canonical text, mirroring the existing `bytea` branch. `postgres-interval` only sets **non-zero** components, so the generic object copier produced `{days:1}` for one row and `{hours:2}` for the next, dropped the prototype's `toPostgres()`/`toISOString()`, and left a consumer reading `.seconds` with `undefined` rather than `0`.

## Tests

Five, two failing before. The other three guard the boundaries and would have caught a sloppier fix: a `text` column that looks numeric, a value that can't survive a double, and `bulkParse` called with no descriptors at all.

## Verification

8 Postgres suites, 77 tests. The two Neo4j failures in a full-suite run pass in isolation (93s vs normal timings) — the usual contention, and this change is Postgres-only.

## Summary by CodeRabbit

* **Bug Fixes**
  * PostgreSQL `int8` and `numeric` values are now converted to numbers when safe, while preserving large values as text to avoid precision loss.
  * Numeric-looking values in text columns remain unchanged as text.
  * PostgreSQL interval values now display as readable text, such as “1 day.”
  * Query result parsing now consistently respects column type information, including when processing multiple rows.

### DIFF
--- a/connection/__tests__/postgresql/postgres-parser.ts
+++ b/connection/__tests__/postgresql/postgres-parser.ts
@@ -154,3 +154,81 @@ describe("PostgreSQL Record Parser", () => {
     expect((result.timestamps as unknown[])[0]).toBeInstanceOf(Date);
   });
 });
+
+describe("PostgresRecordParser — numeric promotion and interval (#1307)", () => {
+  const parser = new PostgresRecordParser();
+
+  // pg-types registers no parser for OID 20 (int8) or 1700 (numeric), so both
+  // arrive as text. SELECT count(*) therefore yielded the STRING "42" on
+  // PostgreSQL and the NUMBER 42 on Neo4j for the same logical query.
+  const INT8 = 20;
+  const NUMERIC = 1700;
+  const TEXT = 25;
+
+  const fields = (defs: Array<[string, number]>) =>
+    defs.map(([name, dataTypeID]) => ({ name, dataTypeID }));
+
+  it("promotes int8 and numeric text to numbers", () => {
+    const [rec] = parser.bulkParse(
+      [{ total: "42", amount: "-12.50" }],
+      fields([
+        ["total", INT8],
+        ["amount", NUMERIC],
+      ]),
+    );
+    expect(rec.toObject().total).toBe(42);
+    expect(rec.toObject().amount).toBe(-12.5);
+  });
+
+  it("leaves a text column alone even when it looks numeric", () => {
+    // This is why promotion is gated on the column's OID rather than on the
+    // string's shape: a product code of "42" must stay a string.
+    const [rec] = parser.bulkParse([{ sku: "42" }], fields([["sku", TEXT]]));
+    expect(rec.toObject().sku).toBe("42");
+  });
+
+  it("keeps the string when Number() would not round-trip exactly", () => {
+    // Same contract as the Neo4j side's inSafeRange() check: precision beats
+    // type-consistency. 9007199254740993 is 2^53+1 and cannot survive a
+    // double, so silently returning 9007199254740992 would be data loss.
+    const [rec] = parser.bulkParse(
+      [{ big: "9007199254740993", money: "1.10" }],
+      fields([
+        ["big", INT8],
+        ["money", NUMERIC],
+      ]),
+    );
+    // Real data loss: 2^53+1 would silently become ...992.
+    expect(rec.toObject().big).toBe("9007199254740993");
+    // NOT data loss: a trailing zero is display formatting, and refusing to
+    // promote here is what leaves money columns sorting as text.
+    expect(rec.toObject().money).toBe(1.1);
+  });
+
+  it("emits interval as text rather than a sparse object", () => {
+    // postgres-interval only sets the components that are non-zero, so the
+    // key set changed row to row: INTERVAL '1 day' gave {days:1} while
+    // INTERVAL '2 hours' gave {hours:2}. A consumer reading .seconds got
+    // undefined for most rows instead of 0.
+    const interval = Object.create({
+      toPostgres() {
+        return "1 day";
+      },
+    });
+    interval.days = 1;
+
+    const [rec] = parser.bulkParse(
+      [{ span: interval }],
+      fields([["span", 1186]]),
+    );
+    expect(typeof rec.toObject().span).toBe("string");
+    expect(rec.toObject().span).toBe("1 day");
+  });
+
+  it("still parses when no field descriptors are supplied", () => {
+    // bulkParse(rows) without fields is used by callers that have no
+    // descriptors; it must not throw, it simply cannot promote.
+    const [rec] = parser.bulkParse([{ total: "42" }]);
+    expect(rec.toObject().total).toBe("42");
+  });
+});
--- a/connection/src/postgresql/PostgresConnectionModule.ts
+++ b/connection/src/postgresql/PostgresConnectionModule.ts
@@ -213,13 +213,16 @@ export class PostgresConnectionModule extends ConnectionModule {
 
       // Parse results to NeodashRecord format
       const parsedRecords = config.parseToNeodashRecord
-        ? this.parser.bulkParse(limitedRows)
+        ? this.parser.bulkParse(limitedRows, fields)
         : limitedRows;
 
       // Set fields if callback is provided
       if (callbacks.setFields) {
         if (limitedRows.length > 0 && config.parseToNeodashRecord) {
-          const firstRecord = this.parser.bulkParse([limitedRows[0]])[0];
+          const firstRecord = this.parser.bulkParse(
+            [limitedRows[0]],
+            fields,
+          )[0];
           callbacks.setFields(
             firstRecord.getFields(config.useNodePropsAsFields),
           );
--- a/connection/src/postgresql/PostgresRecordParser.ts
+++ b/connection/src/postgresql/PostgresRecordParser.ts
@@ -5,13 +5,43 @@ import { NeodashRecord } from "@neoboard/connector-sdk";
  * PostgreSQL Record Parser
  * Converts PostgreSQL result sets to NeodashRecord format.
  */
+/** pg type OIDs that arrive as TEXT because pg-types registers no parser. */
+const OID_INT8 = 20;
+const OID_NUMERIC = 1700;
+
 export class PostgresRecordParser extends NeodashRecordParser {
+  /**
+   * Parses rows, optionally promoting int8/numeric columns to numbers (#1307).
+   *
+   * `fields` is passed per call rather than stored on the instance: the parser
+   * is constructed once and shared across every concurrent query the scheduler
+   * dispatches on a connection, so per-query state here would corrupt across
+   * queries.
+   */
+  bulkParse(
+    records: Record<string, unknown>[],
+    fields?: ReadonlyArray<{ name: string; dataTypeID: number }>,
+  ): NeodashRecord[] {
+    const numericColumns = new Set(
+      (fields ?? [])
+        .filter(
+          (f) => f.dataTypeID === OID_INT8 || f.dataTypeID === OID_NUMERIC,
+        )
+        .map((f) => f.name),
+    );
+    return records.map((r) => this._parse(r, numericColumns));
+  }
+
   /**
    * Parses a single PostgreSQL row into a NeodashRecord.
    * @param _record - A single row from PostgreSQL query results
+   * @param _numericColumns - column names whose text should become numbers
    * @returns A NeodashRecord instance
    */
-  _parse(_record: Record<string, unknown>): NeodashRecord {
+  _parse(
+    _record: Record<string, unknown>,
+    _numericColumns?: ReadonlySet<string>,
+  ): NeodashRecord {
     // If already a NeodashRecord, return as is
     if (_record instanceof NeodashRecord) {
       return _record;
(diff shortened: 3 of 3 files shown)


alfredo1996:
Two things — a process slip on my side, and a decision on the finding.

**The slip:** I merged this while your comment was unread. I checked that a review existed and that there was one comment, then merged in the same step instead of reading it first. That's on me; the evaluation below should have happened before the merge, not after.

**The finding — declining, with reasoning.** You're right about the mechanism: `Number("1000000000000000000000")` is `1e21`, whose `String()` is `"1e+21"`, so `canonicalNumeric` doesn't match and the value stays a string despite round-tripping exactly.

But normalising exponent notation would make the typing *inconsistent within a column*, which I think is worse than a missed promotion:

```
1000000000000000000000 -> 1e+21 | exact round-trip: true
1000000000000000000001 -> 1e+21 | exact round-trip: false
```

Two adjacent values in the same `numeric` column: the first would promote to a number, the second would stay a string. A consumer doing `typeof v === "number"` — which is exactly what `export-utils.ts` does, and the reason this issue exists — would then branch differently row to row within one column.

Above 2^53 the current behaviour fails **closed** and uniformly, which matches the contract the Neo4j side already implements: `inSafeRange()` switches to BigInt beyond the safe range regardless of whether a particular value happens to be representable. Since #1307's whole purpose was making the two connectors agree, diverging here would undercut it.

So: real observation, deliberate no-change. If large-numeric promotion is wanted later, the honest version is a safe-range check (`Number.isSafeInteger`) rather than exponent normalisation — same rule for every row, not per-value luck.

PR #1332, opened 2026-07-28
fix(component): guard the raw string, not Number(), when a numeric field is cleared

### LINKED ISSUE #1292: fix(component): clearing a number input writes 0 because the isNaN(Number("")) guard never fires
Two locations, ONE root cause: `Number("")` is `0`, not `NaN`, so every "is this a number?" guard written as `isNaN(Number(raw))` is dead code for the one input that actually reaches it — the empty string. One fix shape applies to both.

## Problem

Emptying a numeric field (select-all → Delete, the normal way to retype a number) silently rewrites the value to `0` instead of leaving the field blank.

`component/src/components/composed/parameter-widgets/number-range-slider.tsx:47-52`:

```tsx
const handleMinInput = (raw: string) => {
  const num = Number(raw);
  if (isNaN(num)) return;                       // never fires for "" — Number("") === 0
  const clamped = Math.min(Math.max(num, min), current[1]);
  onChange([clamped, current[1]]);
};
```

`handleMaxInput` is identical at lines 54-59. Both inputs are fully controlled — `value={current[0]}` (line 86) and `value={current[1]}` (line 97) — so the coerced `0` is written straight back into the field the user just cleared. Per the HTML spec an `<input type="number">` `.value` is either a valid numeric string or `""`, so `""` is the *only* value that can ever reach the guard, and it is exactly the value the guard fails to catch.

User-visible effect on the range slider: clearing the min field snaps the min handle to the axis minimum and drags the max handle down to the old min; this fires `onChange`, which is a parameter change, which re-runs the widget's query with a bound the user never asked for. There is no way to blank the field to retype it.

The same expression sits in `component/src/components/composed/chart-options-panel.tsx:195-196`:

```tsx
value={String(value ?? option.default ?? 0)}
onChange={(e) => onChange(option.key, Number(e.target.value))}
```

Here the coercion is unguarded — clearing any numeric chart option jumps it to `0`.

| File | Line | Expression | Cleared field becomes |
| --- | --- | --- | --- |
| `component/src/components/composed/parameter-widgets/number-range-slider.tsx` | 48-49 | `Number(raw)` + dead `isNaN` guard | `min` (and max handle collapses to old min), re-runs query |
| `component/src/components/composed/parameter-widgets/number-range-slider.tsx` | 55-56 | `Number(raw)` + dead `isNaN` guard | `current[0]`, re-runs query |
| `component/src/components/composed/chart-options-panel.tsx` | 196 | `Number(e.target.value)`, no guard | `0` |

The same pattern exists in `app/` and should be fixed in the same pass, though the blast radius is smaller (config editors, not live query parameters):

| File | Line | Note |
| --- | --- | --- |
| `app/src/components/widget-editor/parameter-config-section.tsx` | 247-283 | `rangeMin` / `rangeMax` / `rangeStep`, all bare `Number(e.target.value)` |
| `app/src/components/widget-editor/advanced-caching-section.tsx` | 38 | `Math.max(1, Number(...))` — clears to `1`, not `0`, but still unclearable |
| `app/src/components/widget-editor/transform-editor.tsx` | 407 | `Math.max(1, Number(...) \|\| 1)` — same |

## Proposed fix

Guard on the raw string before coercing, not on the coerced number. Minimum change:

- `number-range-slider.tsx` — add `if (raw.trim() === "") return;` above `const num = Number(raw);` in both `handleMinInput` and `handleMaxInput`. The now-dead `isNaN` check can stay or go; it is unreachable either way.
- `chart-options-panel.tsx:196` — skip the callback on empty: `onChange={(e) => { if (e.target.value === "") return; onChange(option.key, Number(e.target.value)); }}`.

This is the smallest change that stops the stored value being destroyed. It does not make the field *visibly* blank while typing (the controlled `value` still renders the last committed number); accepting a transient empty string requires local input state and is out of scope for this issue.

## Required tests

Component layer, vitest + @testing-library/react (jsdom).

1. `component/src/components/composed/__tests__/parameter-widgets.test.tsx` — extend the existing `describe("NumberRangeSlider")` block (starts line 901). Render with `min={0} max={1000} value={[200, 800]}`, `fireEvent.change` the `price minimum` input with `{ target: { value: "" } }`, assert `onChange` was **not** called. Fails today: `onChange` is called with `[0, 800]`. Repeat for the `price maximum` input — fails today with `[200, 200]`. Add a companion assertion that a real edit (`"300"`) still calls `onChange` with `[300, 800]`, so the guard cannot be "fixed" by disabling the handler.
2. `component/src/components/composed/__tests__/chart-options-panel.test.tsx` — for a `number`-typed option, clear the input and assert the panel's `onChange` was not called with `(option.key, 0)`. Fails today: it is.
3. E2E, `app/e2e/parameter-types.spec.ts` — on a dashboard with a number-range parameter widget, clear the min input and assert the widget does not issue a new query and the displayed range is unchanged. Guards the user-visible symptom (spurious re-query on an in-progress edit) that the jsdom tests only cover at the callback level.

## Acceptance criteria

- [ ] Clearing the min or max input in `NumberRangeSlider` does not call `onChange` and does not move either slider handle.
- [ ] Clearing a `number` chart option in `chart-options-panel.tsx` does not commit `0`.
- [ ] Typing a valid number in any of those fields still commits, clamped to `[min, max]` as before.
- [ ] The three `app/src/components/widget-editor/` sites listed above no longer coerce an empty field to a value.
- [ ] The two new component tests fail on `dev` and pass on the branch.
- [ ] `app/e2e/parameter-types.spec.ts` covers the cleared-min case and passes.
- [ ] No remaining `isNaN(Number(` guard in the repo that relies on `""` producing `NaN`.

alfredo1996:
Closes #1292

## Problem

`Number("")` is `0`, not `NaN` — so every guard written as `isNaN(Number(raw))` is **dead code for the one input that can actually reach it**. Per the HTML spec an `<input type="number">.value` is either a valid numeric string or `""`, so the empty string is the only value the guard was ever meant to catch, and the only one it misses.

Emptying a field — select-all → Delete, the normal way to retype a number — therefore silently committed a value the user never typed.

On the range slider both inputs are fully controlled, so the coerced `0` was written straight back into the field just cleared: the min handle snapped to the axis minimum, the max handle collapsed to the old min, and `onChange` fired. That's a **parameter change**, which re-runs the widget's query with a bound nobody asked for. There was no way to blank the field at all.

## Six locations, one shape

Two carry live query parameters:

| File | Expression |
|---|---|
| `number-range-slider.tsx` | `handleMinInput` / `handleMaxInput` — dead `isNaN` guard |
| `chart-options-panel.tsx` | unguarded `Number(e.target.value)` |

Four are config editors — smaller blast radius, same trap:

| File | Note |
|---|---|
| `parameter-config-section.tsx` | `rangeMin` / `rangeMax` / `rangeStep` |
| `advanced-caching-section.tsx` | `Math.max(1, Number(""))` is `1`, not `0` — **unclearable** rather than zeroed |
| `transform-editor.tsx` | `Math.max(1, Number("") \|\| 1)`, same |

The dead `isNaN` checks are deliberately left in place: they're unreachable either way, and removing them invites someone to reinstate the same mistake later.

## Tests

Three, two failing before. The third asserts a real edit still commits `[300, 800]` — so the guard **cannot be "fixed" by disabling the handler**, which is exactly what the first two would otherwise reward. That failure mode is worth guarding against explicitly; a test that only asserts "onChange was not called" is trivially satisfiable the wrong way.

## Deliberately not addressed

The field still renders the last committed number while typing, rather than appearing blank. Accepting a transient empty string requires local input state — a different change, and one that alters typing behaviour for every numeric input in the product. This PR is the smallest change that stops the stored value being destroyed.

## Verification

2025 component tests, 268 widget-editor tests, both packages typecheck.

## Summary by CodeRabbit

* **Bug Fixes**
  * Numeric editor fields can now be cleared without automatically committing unintended values such as 0 or 1.
  * Fixed range, caching, transform, chart option, and number-range controls to preserve valid values while users edit them.
  * Numeric changes continue to apply correctly when a valid value is entered.

* **Tests**
  * Added coverage confirming cleared range inputs do not trigger unintended updates and valid edits still apply.

### DIFF
--- a/app/src/components/widget-editor/__tests__/advanced-caching-section.test.tsx
+++ b/app/src/components/widget-editor/__tests__/advanced-caching-section.test.tsx
@@ -109,6 +109,17 @@ describe("AdvancedCachingSection", () => {
     expect(mockSetCacheTtlMinutes).toHaveBeenCalledWith(1);
   });
 
+  it("ignores a cleared TTL instead of committing 1 (#1292)", () => {
+    mockEnableCache = true;
+    render(<AdvancedCachingSection />);
+    fireEvent.change(screen.getByTestId("cache-ttl"), {
+      target: { value: "" },
+    });
+    // Math.max(1, Number("")) is 1, so clearing the field silently committed
+    // 1 and made the input unclearable.
+    expect(mockSetCacheTtlMinutes).not.toHaveBeenCalled();
+  });
+
   it("pluralizes minutes correctly", () => {
     mockEnableCache = true;
     mockCacheTtlMinutes = 1;
--- a/app/src/components/widget-editor/__tests__/parameter-config-section-ui.test.tsx
+++ b/app/src/components/widget-editor/__tests__/parameter-config-section-ui.test.tsx
@@ -389,22 +389,33 @@ describe("ParameterConfigSection", () => {
     expect(mockSetChartOptions).toHaveBeenCalled();
   });
 
-  it("ignores a cleared range input instead of committing 0 (#1292)", () => {
-    mockStoreState.paramUIType = "number-range";
-    mockStoreState.chartOptions = { rangeMin: 0, rangeMax: 100, rangeStep: 1 };
-    render(
-      <ParameterConfigSection
-        seedQueryExecution={baseSeedExecution}
-        seedPreviewOptions={null}
-      />,
-    );
-    fireEvent.change(screen.getByLabelText("Range maximum"), {
-      target: { value: "" },
-    });
-    // Number("") is 0 — clearing the field used to write rangeMax: 0, which
-    // the controlled input then rendered back, so it could not be retyped.
-    expect(mockSetChartOptions).not.toHaveBeenCalled();
-  });
+  it.each(["Range minimum", "Range maximum", "Range step"])(
+    "ignores a cleared %s instead of committing 0 (#1292)",
+    (label) => {
+      mockStoreState.paramUIType = "number-range";
+      mockStoreState.chartOptions = {
+        rangeMin: 0,
+        rangeMax: 100,
+        rangeStep: 1,
+      };
+      render(
+        <ParameterConfigSection
+          seedQueryExecution={baseSeedExecution}
+          seedPreviewOptions={null}
+        />,
+      );
+      fireEvent.change(screen.getByLabelText(label), { target: { value: "" } });
+      // Number("") is 0 — clearing the field used to write 0, which the
+      // controlled input rendered straight back, so it could not be retyped.
+      expect(mockSetChartOptions).not.toHaveBeenCalled();
+
+      // The guard must not swallow real edits.
+      fireEvent.change(screen.getByLabelText(label), {
+        target: { value: "7" },
+      });
+      expect(mockSetChartOptions).toHaveBeenCalled();
+    },
+  );
 
   // ── cascading editor (regression: #861) ────────────────────────────
   it("shows parent-parameter input for cascading type", () => {
--- a/app/src/components/widget-editor/__tests__/parameter-config-section-ui.test.tsx
+++ b/app/src/components/widget-editor/__tests__/parameter-config-section-ui.test.tsx
@@ -389,6 +389,23 @@ describe("ParameterConfigSection", () => {
     expect(mockSetChartOptions).toHaveBeenCalled();
   });
 
+  it("ignores a cleared range input instead of committing 0 (#1292)", () => {
+    mockStoreState.paramUIType = "number-range";
+    mockStoreState.chartOptions = { rangeMin: 0, rangeMax: 100, rangeStep: 1 };
+    render(
+      <ParameterConfigSection
+        seedQueryExecution={baseSeedExecution}
+        seedPreviewOptions={null}
+      />,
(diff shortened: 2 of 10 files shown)


alfredo1996:
Declining the E2E, with the reasoning.

`NumberRangeSlider` is a pure presentational component in `component/` — its only output is the `onChange` tuple (`number-range-slider.tsx:26`, "Pure presentational"). Everything the finding asks an E2E to observe is downstream of that one call: no `onChange` → no parameter-store write → no refetch → no change to the displayed range. The component test asserting `onChange` is not called *is* the assertion that no query fires; an E2E would re-derive it through three layers and one browser.

The package boundary makes it structural rather than a judgement call: `component/` has no stores and no API calls, so there is no "owning flow" inside this package to test, and the app-side guards each have their own jsdom test in this PR (`chart-options-panel`, `advanced-caching-section`, `parameter-config-section`, `transform-editor`).

The finding did point at something real, though: my first pass only tested the path the guard blocks. All six cases now assert both halves — cleared input ignored **and** a real value still commits — so a guard that blocked everything would fail. That was the actual hole.

PR #1344, opened 2026-07-28
feat(ops): check that ENCRYPTION_KEY can actually decrypt what is stored (#1274)

### LINKED ISSUE #1274: fix(ops): nothing verifies ENCRYPTION_KEY can decrypt stored credentials — wrong key boots clean, then every widget fails
## Problem

Every layer checks that `ENCRYPTION_KEY` is **well-formed**. Nothing checks that it is the **right key** — the one that encrypted the credentials already in the database.

- `app/src/lib/env-config.ts` validates the 64-hex shape, and `/api/health` reports it as `set`.
- `neoboard doctor` checks Docker, Node, ports, `node_modules`, and whether `.env.local` exists — it never looks at the key at all.
- `crypto.ts` only discovers the mismatch at **use** time, and throws Node's raw AES-GCM error: `Unsupported state or unable to authenticate data`.

So an instance with a regenerated or mismatched key boots clean, passes health, passes doctor, and then fails on **every widget** with an error that names neither the key nor the fix. The admin sees "all my dashboards broke" and has no thread to pull.

This is not hypothetical: the same trap was hit locally when a demo re-seed ran with the host's `app/.env.local` key against containers holding `docker/.env`'s key — every widget failed to decrypt, and the cause took a manual diff of two env files to find.

## Why it matters for operators

`ENCRYPTION_KEY` is the one value the docs correctly say is unrecoverable if lost. That makes "is this the right key?" the single highest-value operational check in the product, and it is the one check nobody runs. It is also the failure mode most likely to hit during exactly the operations we tell admins to do: restoring a backup, migrating hosts, rotating secrets, or standing up a second environment against an existing database.

## Proposed fix

One probe, reused in both places: read the oldest stored `connections.configEncrypted`, attempt `decrypt`, and report the outcome. Cheap (one row), read-only, and conclusive.

- **`neoboard doctor`** — a `Credential decryption` check: `ok` when a stored credential decrypts, `fail` with "ENCRYPTION_KEY does not match the stored credentials — this is the key from <source>; the data was encrypted with a different one" when it doesn't, and `skip` when there are no connections yet.
- **`/api/health`** (admin view only, alongside the existing migration status) — the same boolean, so a probe or an ops dashboard can alert on it. It must stay out of the unauthenticated payload: "this instance cannot decrypt its own secrets" is a gift to an attacker.
- The error thrown at query time should also name the likely cause instead of surfacing the raw AES-GCM message.

Explicitly **not** proposed: storing a key fingerprint or checksum anywhere. That reintroduces a key-derived artifact at rest, which the direct-key decision deliberately avoids.

## Required tests

**Unit (`cli/src/__tests__/`)**
- the doctor check reports `ok` when the probe decrypts, `fail` when `decrypt` throws, and `skip` when no connections exist
- the failure message names `ENCRYPTION_KEY` and does not include any ciphertext or key material

**API route (`app/src/app/api/health/__tests__/`)**
- an admin session sees the decryption status; an anonymous probe does not
- a decrypt failure degrades the payload without turning the health probe into a 503 (the app is up; its secrets are unreadable — those are different alarms)

**Integration (`connection/` or a CLI integration test)**
- encrypt with key A, restart the probe with key B, assert the check fails — the actual regression this prevents
- with `ENCRYPTION_KEY_OLD` set to key A, the check passes, since rotation is a supported state

## Acceptance criteria

- [ ] `neoboard doctor` fails loudly when the configured key cannot decrypt stored credentials
- [ ] Admin-only health output carries the same signal
- [ ] Query-time decrypt failures name `ENCRYPTION_KEY` as the likely cause
- [ ] No key material, fingerprint, or ciphertext is logged or persisted by any of the above
- [ ] Key rotation via `ENCRYPTION_KEY_OLD` still reports healthy

alfredo1996:
## What

Every layer checked the key was well-**formed** — `env-config` validates 64 hex chars, `/api/health` reports it `set`, `doctor` never looked at it at all. Nothing checked it was the **right** key.

So an instance with a regenerated or mismatched key booted clean, passed health, passed doctor, and then failed on *every widget* with Node's raw `Unsupported state or unable to authenticate data` — an error naming neither the key nor the fix. This is the failure mode most likely to hit during exactly the operations the docs tell admins to perform: restoring a backup, migrating hosts, rotating secrets, standing a second environment up against an existing database.

## Three surfaces, one probe

Read one stored ciphertext, try to decrypt, report the outcome.

| Surface | What it gains |
| --- | --- |
| `neoboard doctor` | a `Credential decryption` check |
| `/api/health` | admin-only `credentials.decryption` — "this instance cannot decrypt its own secrets" is a gift to an attacker |
| `decrypt()` itself | names `ENCRYPTION_KEY` instead of re-throwing Node's message |

**Four outcomes, not two.** `no-credentials` and `unavailable` both report **skip**, not ok. Claiming ok when nothing was verified is the same false confidence this check exists to remove; and reporting an unreachable database as a key mismatch would send an operator to rotate a key that's fine. That needed a `skip` status on `CheckResult`, which had only ok/warn/fail.

Health degrades to `unknown` rather than throwing, and a mismatch does **not** become a 503 — the app is up, its secrets are unreadable, and those are different alarms. Conflating them pulls an instance out of a load balancer for a problem no restart fixes.

Deliberately **not** a stored key fingerprint, per the issue: that puts a key-derived artifact at rest, which the direct-key design avoids on purpose.

## Two things caught while building it

**The first `decrypt()` message broke an existing test, correctly.** Malformed ciphertext (wrong segment count) is *not* a key problem, and reporting it as one is the exact misdiagnosis described above. Format errors now re-throw untouched.

**The existing rotation test asserted `expect(() => decrypt(ct)).toThrow()`** — which passes on the very message this issue is about. Tightened to name `ENCRYPTION_KEY`, plus a case asserting the message leaks neither key nor ciphertext nor plaintext. An error message is the one place secrets reliably end up in logs.

## Verification beyond the 25 new unit tests

- **Cross-implementation.** The CLI can't import `app/` (package boundary), so its crypto mirrors `crypto.ts`. Fed it real ciphertext from the app's own `encrypt()`: reads with the right key, rejects the wrong one.
- **Live.** A real Postgres, a real `connection` row, through the real `psql` path, no mocks:
  ```
  key A (the right key)        -> ok
  key B (well-formed, wrong)   -> mismatch
  malformed key                -> unavailable
  no connection table          -> unavailable
  ```
- **The mismatch tests were confirmed to FAIL when the auth-tag check is removed.** A crypto test that passes on a broken cipher tests nothing.

Closes #1274

### DIFF
--- a/app/src/app/api/health/__tests__/route.test.ts
+++ b/app/src/app/api/health/__tests__/route.test.ts
@@ -5,6 +5,7 @@ const mockValidate = vi.fn();
 const mockDbExecute = vi.fn();
 const mockRequireSession = vi.fn();
 const mockListSchedulers = vi.fn();
+const mockProbe = vi.fn();
 
 vi.mock("next/server", () => nextResponseMockFactory());
 vi.mock("@/lib/env-config", () => ({
@@ -21,6 +22,9 @@ vi.mock("@/lib/auth/session", () => ({
 }));
 vi.mock("@/lib/query/scheduler-registry", () => ({
   listSchedulers: mockListSchedulers,
+}));
+vi.mock("@/lib/crypto/credential-health", () => ({
+  probeCredentialDecryption: mockProbe,
 }));
 
 describe("GET /api/health", () => {
@@ -47,8 +51,12 @@ describe("GET /api/health", () => {
     vi.doMock("@/lib/query/scheduler-registry", () => ({
       listSchedulers: mockListSchedulers,
     }));
+    vi.doMock("@/lib/crypto/credential-health", () => ({
+      probeCredentialDecryption: mockProbe,
+    }));
     mockRequireSession.mockRejectedValue(new Error("unauthenticated"));
     mockListSchedulers.mockReturnValue([]);
+    mockProbe.mockResolvedValue("ok");
     const mod = await import("../route");
     GET = mod.GET;
   });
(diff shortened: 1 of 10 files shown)


alfredo1996:
Sonar failed at **79.5%** new-code coverage. `app/src/lib/crypto/credential-health.ts` was at **0%** — the health route test mocks it wholesale, so the module never executed. A mock made the code look tested, which is the exact trap I've been writing up all session.

Writing the real tests immediately surfaced a defect the mocked version could not have: an **unset** `ENCRYPTION_KEY` reported `mismatch`, because `getKey()` throws inside the inner try/catch. That sends an operator to rotate a key that does not exist — the same misdiagnosis as calling malformed ciphertext a key failure, which I'd fixed one function over an hour earlier. It now returns `unknown` and skips the query, matching the CLI probe.

Nine cases, all against real AES-GCM ciphertext rather than fixtures: right key, wrong-but-well-formed key, mid-rotation via `ENCRYPTION_KEY_OLD` (→ `ok`, a supported state), tampered ciphertext, empty result, empty column, throwing query, unset key, malformed key, no plaintext in the return.

Plus doctor's three uncovered `readEncryptionKey` branches — `docker/.env` vs `app/.env.local`, absent file, unparseable file. The last matters: a truncated `.env` must not crash the command an operator runs when things are already broken.

`src/lib/crypto` is now at 100% lines. 3181 app tests, 331 CLI tests, lint clean.

Issue #991, opened 2026-06-10
[P2] cli: portability bugs — Windows infinite loop, unquoted paths, curl/pg_isready PATH assumptions


alfredo1996:
**From** the 2026-06-10 package audit (`claude_code_docs/review-2026-06-package-audit.md`, report-only section — promoting to an issue).

## Problems

1. **Windows infinite loop**: `findProjectRoot` only terminates at `"/"` (`cli/src/lib/config.ts:25`); on Windows `dirname("C:\\") === "C:\\"`, so a run outside a repo loops forever.
2. **Paths with spaces break**: compose file paths are interpolated unquoted into shell strings (`docker compose -f ${file}`, `cli/src/lib/docker.ts:22,28,39`).
3. **PATH assumptions**: local-mode health checks shell out to `curl` (`docker.ts:107-110,126-130`; `status.ts:8-10`) and `pg_isready`/`psql`/`pg_dump` — none guaranteed on PATH, no actionable error when missing.

## Suggested fix

Quote interpolated paths, replace `curl` probes with Node `fetch`, bound the root-finding loop, and either support Windows or fail fast with a clear "use WSL" message.

alfredo1996:
Windows findProjectRoot loop fixed, compose paths quoted — fixed in #1031 (merged to release/1.1). CLI suite 278 + real-Docker integration 9/9.

PR #1345, opened 2026-07-28
fix(cli): make the published package work outside the monorepo (#1315)

### LINKED ISSUE #1315: fix(cli): published @neoboard/cli cannot run outside the monorepo — root detection throws and compose files aren't packaged
## Problem

`.github/workflows/release.yml:101-142` publishes `@neoboard/cli` to npm with provenance on every `v*` tag. The moment a tag is cut, `npx @neoboard/cli setup` becomes a thing strangers will run — and it will fail immediately, for two independent reasons.

**1. Root detection requires a monorepo checkout.** `cli/src/lib/config.ts:23-44`:

```ts
export function findProjectRoot(startDir?: string): string {
  let dir = startDir ?? dirname(fileURLToPath(import.meta.url));
  for (;;) {
    const pkg = JSON.parse(readFileSync(join(dir, "package.json"), "utf-8"));
    if (pkg.name === "neoboard") return dir;
    ...
  }
  throw new Error("Could not find NeoBoard project root (package.json with name 'neoboard')");
}
```

It walks up from the CLI's **own install location** looking for the monorepo root. Under `npx`, that location is an npm cache directory with no `neoboard` package.json above it, so every command that touches `paths.root` throws before doing anything.

**2. The package does not ship what the CLI needs.** `cli/package.json` declares `files: ["dist", "README.md", "LICENSE"]`. The compose files under `docker/` — which `composeFile()` resolves via `paths.dockerDir` — are not included. Even with root detection fixed, there would be nothing to start.

The README already promises this path: *"an `npx @neoboard/cli` standalone install path is on the roadmap but the package is not on npm yet"* (`README.md:50`). The publish step is wired; only the tag is missing. So today's state is that the first tag ships a package whose primary command cannot run.

## Relationship to #1314

The one-line shell installer deliberately does **not** depend on the CLI — it drives Docker directly, so it is unblocked by this. This issue is the Node-native alternative for people who prefer `npx`, and it can follow.

## Proposed fix

**Root resolution needs a mode, not a search.** Two situations, and the CLI cannot tell them apart today:

- *inside the monorepo* — behave as now (contributors)
- *installed standalone* — there is no project root. Use a working directory (`$NEOBOARD_DIR`, else `./neoboard`, else `~/.neoboard`), create it, and treat it as the root for config, `.env` and compose.

Make `findProjectRoot()` return `null` rather than throw, and have the caller decide. A thrown error deep in a path helper is why this surfaces as an unrelated-looking crash.

**Ship the compose files** — add `docker/` (or a pruned subset) to `files`, and prefer `docker-compose.prod-full.yml`, which pulls `ghcr.io/...` rather than building from source (`docker-compose.prod-full.yml:103`). A standalone user has no source to build.

**Fail honestly if a command genuinely needs the repo.** `neoboard dev`, `db seed` and the plugin commands assume a checkout. In standalone mode they should say so, not throw a root-detection error.

## Required tests

**Unit — `cli/src/__tests__/lib/`**
- `findProjectRoot` returns null (not throws) when no `neoboard` package.json exists above the CLI
- standalone mode resolves config/env/compose under the working directory; `NEOBOARD_DIR` overrides
- monorepo mode is unchanged — this is the regression risk, since every existing command depends on it

**Packaging — a test that fails if the package would be broken:**
- assert every path `composeFile()` can return exists inside the packaged `files` set. A pure unit test that reads `cli/package.json` and checks the compose paths would have caught this before publish.

**Integration**
- `npm pack` the CLI, install the tarball into a temp dir outside the repo, run `neoboard setup --help` and one real command; assert no root-detection error. This is the only test that reproduces the actual failure.

## Acceptance criteria

- [ ] A tarball-installed CLI runs outside the monorepo without a root-detection error
- [ ] The published package contains the compose files it references
- [ ] Standalone mode uses a documented working directory
- [ ] Repo-only commands fail with a clear explanation
- [ ] Monorepo behaviour is unchanged for contributors
- [ ] A packaging test fails if a referenced file is not shipped

alfredo1996:
## What

`release.yml` publishes `@neoboard/cli` on every `v*` tag, so the first tag makes `npx @neoboard/cli setup` real for strangers. It failed immediately, twice over.

**1. Root detection required a checkout.** `findProjectRoot` walked up from the CLI's own install location looking for the monorepo; under `npx` that's an npm cache directory, and it **threw**. A throw that deep in a path helper surfaced as an unrelated-looking crash — `setup` died in root detection, reporting nothing about standalone mode.

It now returns `null`, and `resolveRoot` decides: the checkout when there is one, else `$NEOBOARD_DIR`, else `./neoboard`, created on demand. **The monorepo always wins**, so a stray `NEOBOARD_DIR` can't silently redirect a contributor's checkout.

**2. The package shipped nothing to start.**

## The part worth reading

Adding `"docker"` to `files` **does nothing**. npm resolves `files` relative to the *package* root; `docker/` lives one level up at the *repo* root. `npm pack` succeeds and ships no compose file, silently.

That matters more than the fix, because **my first packaging test asserted `pkg.files.includes("docker")` — and passed.** It proved nothing and would have let this ship a second time. The test now runs `npm pack --dry-run --json` and asserts on what would actually be published. A `prepack` step stages the four assets a standalone install can use into `cli/docker/` (gitignored, removed by `postpack`).

## Commands that genuinely need the source now say so

`dev`, `db seed` and the plugin commands run `scripts/` tooling that's deliberately **not** shipped — 568K to make one command work is the wrong trade. `assertCheckout` names the command, gives the clone commands, and lists what *does* work standalone.

## Errors no longer surface as stack traces

`program.parse()` left a rejected action to Node's unhandled-rejection dump — the message buried under a trace rooted in `dist/`, reading as a crash rather than as the CLI telling you something. `parseAsync().catch` prints the message and exits 1.

## Two existing tests changed contract

Both asserted the throw that is now a `null`. The #991 one still proves what it was written for — a non-terminating loop would *hang* rather than return null — so only the assertion changed, and the comment says so.

## Verified against a real packed tarball, installed outside the monorepo

```
neoboard --help                works
neoboard config list           works, creates ./neoboard/
NEOBOARD_DIR=… config list     honours the override
neoboard status                exit 0
neoboard dev                   clean message, exit 1, no stack trace
node cli/dist/index.js dev     in-monorepo path unaffected
tarball contains docker/docker-compose.prod-full.yml
```

Also fixes the prepack script logging to **stdout**, which corrupts `npm pack --json` for every consumer, not just this test.

344 CLI unit tests pass; lint clean. (`integration/demo-flow.test.ts` needs a running stack and fails identically on a stashed tree.)

Closes #1315

## Summary by CodeRabbit

- **New Features**
  - CLI packages now include the required Docker Compose assets for standalone installations.
  - Standalone CLI installs can automatically use a local working directory without a full repository checkout.
  - Added clearer validation for commands that require a repository checkout.

- **Bug Fixes**
  - CLI command failures now display clean, user-friendly error messages instead of unhandled stack traces.
  - Improved project-root detection when running outside a repository checkout.

- **Tests**
  - Added coverage for standalone installation behavior, package contents, and checkout validation.


alfredo1996:
CI caught a real gap my local run missed: the packaging test asserts on the **actual** `npm pack` output, and CI has no `cli/dist`, so the tarball genuinely lacked `dist/index.js`.

The test was right. A package packed without a build **is** broken — the release workflow only escapes it because it runs an explicit `npm run build` step first. `prepack` now builds, which is what prepack is for, so `npm pack` produces a complete package for anyone rather than only for the one workflow that happens to build first. `prepublishOnly`'s build is redundant now and removed.

Also replaced `postpack: rm -rf docker`. That path is relative to whatever directory npm runs it from, and one level up is the repo's real `docker/`. Both paths now come from `import.meta.url`, so cwd cannot make it delete the wrong tree.

Verified from a clean `dist` with the exact invocation `release.yml` uses — `npm publish --workspace=@neoboard/cli --dry-run` stages the compose files, ships `dist/index.js`, and removes `cli/docker` afterwards.

**One thing I checked and dismissed**, since it looks alarming in the dry-run output: `npm warn publish "bin[neoboard]" script name dist/index.js was invalid and removed`. It is pre-existing on `release/1.4` and harmless — the packed manifest keeps `bin`, and installing the tarball produces a working `node_modules/.bin/neoboard` (`neoboard --version` → `1.0.0`). npm is normalising an in-memory manifest for the dry-run, not the package. I initially read it as "the published CLI has no binary"; that was wrong.

alfredo1996:
All four addressed — every one was valid.

**The Major one was a genuine behavioural gap I had missed.** `resolveRoot`/`isStandalone` searched from `import.meta.url` only, so a globally installed or npx'd CLI run **inside** a checkout was treated as standalone: `dev` and `db seed` refused, and `config list` would create `./neoboard` inside the user's own source tree with the repo right there. Now searches `process.cwd()` first, falling back to the module path — which still covers the monorepo's own `node cli/dist/index.js` run from an unrelated directory.

Verified with the real installed binary, both directions:

| Invoked from | `config get ports.app` | `dev` | `./neoboard` |
|---|---|---|---|
| inside the checkout | 3000 | proceeds | not created |
| a bare temp dir | 3000 | declines, exit 1 | created |

Plus a regression test for the checkout invocation, as suggested.

**The three Minor ones:**

- *"Both tests only assert the returned path, so removing `mkdirSync` would still pass."* — correct, and it is the loose-assertion shape I have been cataloguing all session, in my own test. Now asserts `existsSync`.
- `afterEach` deleted `NEOBOARD_DIR` unconditionally, clobbering a value the developer or CI may have set. Restores it.
- `new URL(...).pathname` keeps the leading slash and percent-encodes spaces, so a checkout under `Program Files` would pass a broken `cwd`. `fileURLToPath` instead — this repo has already been bitten by a Windows path bug (#991), so that is a live concern rather than a theoretical one.

346 CLI unit tests pass; lint clean.

PR #1347, opened 2026-07-28
fix(app): tell the user localhost means the container, not their machine (#1346)

### LINKED ISSUE #1346: fix(app): connecting to a host database via localhost fails with an error that never mentions Docker
## Problem

The most common thing a user does after `neoboard demo` is connect their **own** database. On any Docker install that database is on the host, so they type what works everywhere else:

```
neo4j://localhost:7688
```

It fails. The app runs *inside a container*, so `localhost` is the container itself, not the host. Confirmed from `neoboard-app`:

```
localhost:7688              -> FAIL ECONNREFUSED
host.docker.internal:7688   -> OPEN
```

The error the user sees is the driver's:

```
ServiceUnavailable: Could not perform discovery. No routing servers available.
```

`classifyConnectionError` correctly codes that `network`, and the hint reads *"The server is unreachable. Verify the host and port, confirm the database is running, and check that no firewall is blocking the connection."*

Every one of those instructions is a dead end here. The host is right, the port is right, the database **is** running and reachable, and no firewall is involved. The user verifies all four, finds nothing wrong, and has no way to discover that `localhost` means something different inside a container. Nothing in the message, the docs, or the connection form mentions Docker.

## Second half: the fix we'd suggest doesn't work everywhere

`host.docker.internal` resolves automatically on Docker Desktop (macOS/Windows). On **Linux** it does not, unless the container is started with `--add-host=host.docker.internal:host-gateway`.

No compose file sets it:

```
$ grep -n "extra_hosts|host-gateway" docker/*.yml
(no matches)
```

So a hint naming `host.docker.internal` would be correct on a Mac and wrong on a Linux server — worse than no hint, because it sends the user to a name that will not resolve.

## Proposed fix

**1. Classify it.** `classifyConnectionError` receives only the message; give it the URI too. When the error is a network failure **and** the URI host is a loopback address (`localhost`, `127.0.0.1`, `::1`) **and** the app is containerised (`/.dockerenv`), return a distinct code rather than `network`.

The containerisation check matters: in local mode the app runs on the host, where `localhost` is exactly right, and the hint would be actively misleading.

**2. Make the suggestion true on Linux.** Add to the app service in `docker-compose.full.yml`, `docker-compose.prod.yml` and `docker-compose.prod-full.yml`:

```yaml
extra_hosts:
  - "host.docker.internal:host-gateway"
```

Harmless on Docker Desktop, required on Linux.

**3. Document it** on the connections page — one short section on reaching a host database from a Docker install.

## Required tests

**Unit — `app/src/lib/connector/__tests__/`**
- a network-class error with a loopback URI, containerised → the new code; the hint names `host.docker.internal`
- the same error with a loopback URI, **not** containerised → stays `network` (this is the regression that matters: local-mode users must not be told to use a Docker hostname)
- the same error with a non-loopback URI, containerised → stays `network`
- an **auth** error with a loopback URI, containerised → stays `auth_failed`; priority order is unchanged
- `127.0.0.1`, `::1` and `[::1]` all recognised as loopback, not just the literal `localhost`
- a malformed URI does not throw the classifier — it must degrade, since this runs on an error path

**Config — an assertion, not prose**
- every compose file that runs the app service declares `host.docker.internal:host-gateway`. Without this the hint is wrong on Linux, and nothing else would catch it.

**No E2E:** reproducing it needs a database on the host outside the compose network, which the E2E stack has no way to provide.

## Acceptance criteria

- [ ] A loopback connection attempt from a containerised app produces a hint that names Docker and `host.docker.internal`
- [ ] The same attempt from a non-containerised app is unchanged
- [ ] Auth and bad-URI classification are unaffected
- [ ] `host.docker.internal` resolves on Linux from every compose file that runs the app
- [ ] The connections docs cover reaching a host database

alfredo1996:
## What

The most common thing a user does after `neoboard demo` is connect their **own** database. On a Docker install that database is on the host, so they type `neo4j://localhost:7688` — and inside the app container `localhost` is the container.

The driver says `Could not perform discovery. No routing servers available`. That classified as `network`, whose hint told them to *"verify the host and port, confirm the database is running, and check that no firewall is blocking the connection."*

**All four are already correct.** The host is right, the port is right, the database is running and reachable, no firewall involved. Nothing in the message, the docs, or the form mentions Docker. There is no thread to pull.

## How

A new `container_loopback` code, when the error is a network failure **and** the URI host is loopback **and** the app is containerised (`/.dockerenv`).

All three conditions matter:

- **Containerised.** In local mode the app runs on the host, where `localhost` is exactly right — telling that user to use a Docker hostname sends them somewhere that does not exist for them. That regression test is the important one.
- **Network only.** It narrows `network`, never overriding `auth_failed` or `bad_uri`. A loopback auth failure is still an auth failure.
- **Parsed, not substring-matched.** `myhost-localhost.example.com` contains `"localhost"` and is not loopback. An unparseable URI returns false rather than throwing — this runs on an error path, where a throw replaces a bad message with a 500.

## The half that would otherwise have made this a new bug

`host.docker.internal` resolves automatically on Docker Desktop but **not on Linux** without an explicit `host-gateway` mapping — and no compose file had one:

```
$ grep -n "extra_hosts|host-gateway" docker/*.yml
(no matches)
```

A hint naming a hostname that does not resolve is worse than no hint. Added to all three compose files that run the app, with an assertion so its absence can't go unnoticed — proven by deleting it from one file and watching the check name that file.

## Security

The URI reaches the classifier and goes no further; it can carry a password. A test pins that neither the credentials nor the host appear in the returned result.

## Verified against the running stack, before writing any of this

```
from inside neoboard-app:
  localhost:7688             -> FAIL ECONNREFUSED
  host.docker.internal:7688  -> OPEN, 6053 nodes
```

3203 app tests, typecheck, lint clean.

Closes #1346

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in `--expose-host` flag to the start command to map `host.docker.internal` and support connecting to services on the same machine from within containers.

* **Bug Fixes**
  * Improved connection-test error hints for containerized setups that mistakenly use `localhost`/loopback addresses.
  * Enhanced error classification context and ensured user-facing errors avoid revealing sensitive connection details.

* **Tests**
  * Added/expanded automated coverage for container detection, connection error classification, and Docker Compose networking overlays.

### DIFF
--- a/app/src/app/api/connections/[id]/test/route.ts
+++ b/app/src/app/api/connections/[id]/test/route.ts
@@ -11,6 +11,7 @@ import {
   connectionCheckFalseResult,
   connectionTestErrorResult,
 } from "@/lib/connector/connection-test-result";
+import { isContainerised } from "@/lib/connector/is-containerised";
 
 export async function POST(
   _request: Request,
@@ -60,7 +61,12 @@ export async function POST(
         success ? { success: true } : connectionCheckFalseResult(),
       );
     } catch (testError) {
-      return apiSuccess(connectionTestErrorResult(testError));
+      return apiSuccess(
+        connectionTestErrorResult(testError, {
+          uri: credentials.uri,
+          containerised: isContainerised(),
+        }),
+      );
     }
   } catch (error) {
     return handleRouteError(error, "Connection test failed");
--- a/app/src/app/api/connections/test-inline/route.ts
+++ b/app/src/app/api/connections/test-inline/route.ts
@@ -9,6 +9,7 @@ import {
   connectionCheckFalseResult,
   connectionTestErrorResult,
 } from "@/lib/connector/connection-test-result";
+import { isContainerised } from "@/lib/connector/is-containerised";
 
 export async function POST(request: Request) {
   try {
@@ -43,7 +44,12 @@ export async function POST(request: Request) {
         success ? { success: true } : connectionCheckFalseResult(),
       );
     } catch (testError) {
-      return apiSuccess(connectionTestErrorResult(testError));
+      return apiSuccess(
+        connectionTestErrorResult(testError, {
+          uri: config.uri,
+          containerised: isContainerised(),
+        }),
+      );
     }
   } catch (error) {
     return handleRouteError(error, "Connection test failed");
--- a/app/src/lib/__tests__/connector/container-host.test.ts
+++ b/app/src/lib/__tests__/connector/container-host.test.ts
@@ -1,0 +1,65 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+
+const mockIsContainerised = vi.fn();
+vi.mock("@/lib/connector/is-containerised", () => ({
+  isContainerised: () => mockIsContainerised(),
+}));
+
+import { resolveContainerHost } from "@/lib/connector/container-host";
+
+beforeEach(() => vi.clearAllMocks());
+
+describe("resolveContainerHost (#1346)", () => {
+  describe("inside a container", () => {
+    beforeEach(() => mockIsContainerised.mockReturnValue(true));
+
+    it.each([
+      ["neo4j://localhost:7688", "neo4j://host.docker.internal:7688"],
+      ["bolt://127.0.0.1:7687", "bolt://host.docker.internal:7687"],
+      ["neo4j://LOCALHOST:7687", "neo4j://host.docker.internal:7687"],
+    ])("rewrites %s", (input, expected) => {
+      expect(resolveContainerHost(input)).toBe(expected);
+    });
+
+    it("keeps credentials, port, path and query intact", () => {
+      // A rewrite that drops the password or the database name trades one
+      // failure for a more confusing one.
+      expect(
+        resolveContainerHost(
+          "postgresql://u:p%40ss@localhost:5432/app?sslmode=require",
+        ),
+      ).toBe(
+        "postgresql://u:p%40ss@host.docker.internal:5432/app?sslmode=require",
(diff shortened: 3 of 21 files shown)


alfredo1996:
All three addressed.

**Major — `--expose-host` needs Docker full-stack mode.** Real bug I shipped, and worth showing what it actually did. The overlay adds `extra_hosts` to the `neoboard` service, which only the **full** compose defines. Without `--full`, compose refuses the entire project:

```
service "neoboard" has neither an image nor a build context specified: invalid compose project
```

An error about the wrong thing entirely — nothing in it says "you forgot `--full`". Verified against real `docker compose config` both before and after. Local mode is rejected too, with a different message: the app runs on the host there, so `localhost` already reaches the databases and the flag has nothing to do. Three tests, plus the real binary now printing the fix instead of a compose error.

**Minor — assert on the app service, not anywhere in the file.** Correct: the raw substring would also pass from a comment, an unrelated service, or malformed YAML, and a mapping on the *wrong* service is exactly the failure that check exists to catch. Now resolves the project with `docker compose config --format json` and asserts `extra_hosts` on the app service specifically.

**Minor — co-locate tests with sources: declining, with reason.** CLAUDE.md does say tests live next to the file under test, but the `cli/` package centralises **all 30** test files under `cli/src/__tests__/`. Moving one would make that package inconsistent rather than compliant. The convention gap is real and worth settling repo-wide — doing it inside this PR would bury a networking fix under a test migration. Happy to file it separately if you'd like.

341 CLI tests, 42 script tests, lint clean.

Issue #1278, opened 2026-07-27
feat(cli): db dump has no restore counterpart — the one command an admin needs while down


alfredo1996:
## Problem

The CLI can take a backup but cannot restore one. `neoboard db dump` exists (`cli/src/commands/db/dump.ts`); there is no `neoboard db restore`. The command surface is:

`init · setup · start · stop · dev · status · doctor · logs · env · config · db {migrate,reset,seed,dump} · demo · plugin`

Every one of those is "get it running". Nothing covers "run it for a year". Restore is the sharpest instance because the *taking* half is already there, so the asymmetry reads as an oversight rather than a scope decision.

Today an operator restoring a backup must follow a multi-step manual recipe — and per #1219 those recipes do not currently work as written against a non-empty target. So the one procedure an admin performs while their instance is down is the least supported path in the product.

## Why a command rather than better docs

#1219 fixes the prose, which is necessary. But the preconditions it documents are exactly the kind that a human under recovery pressure skips, and each one fails *late* and confusingly:

- the target must be empty or dropped (`--clean --if-exists`), else duplicate-key errors mid-restore
- `MIGRATE_ON_START` must be off until the restore finishes, else the app creates the schema first and collides
- the `ENCRYPTION_KEY` must be the one that was active when the dump was taken, or every credential in the restored data is unreadable — the failure mode in #1274, and it surfaces long after the restore "succeeded"

A command encodes all three as preflight checks, which is worth more than a paragraph telling someone to remember them.

## Proposed scope

`neoboard db restore <file>`, mirroring `db dump`'s docker/local mode handling and its `assertSafeValue` argument guarding:

- refuse to run against a non-empty database unless `--clean` is passed, and say which tables were found
- warn when `MIGRATE_ON_START` is enabled in the resolved env, with the flag to proceed anyway
- after restore, run the #1274 decryption probe and report plainly whether the current key can read the restored credentials
- print what was restored (tables, row counts, dump timestamp) rather than exiting silently

## Required tests

**Unit (`cli/src/__tests__/`)**
- refuses a non-empty target without `--clean`; proceeds with it
- rejects unsafe characters in config-derived values, matching `db dump`'s existing guard
- builds the correct docker vs local invocation for each mode
- surfaces a non-zero exit from the underlying `psql`/`pg_restore` as a failure rather than reporting success

**Integration**
- dump → drop → restore round-trip on a seeded database leaves dashboards and connections intact and decryptable

## Acceptance criteria

[...]

PR #1352, opened 2026-07-28
fix(cli): the credential probe's SQL never survived the shell (#1274)

alfredo1996:
## What

`neoboard doctor` reported **"no stored credentials yet"** on every **local-mode** install — silently disabling the entire check #1274 added this morning.

The local branch built a shell string:

```
psql ... -tAc "SELECT "configEncrypted" FROM connection LIMIT 1"
```

The inner quotes terminated the outer ones, so the shell delivered:

```
psql -tAc SELECT configEncrypted FROM connection LIMIT 1
```

Two failures at once: the statement split across four argv slots, **and** the identifier lost its quoting, so Postgres folded it to `configencrypted`. The column doesn't exist, the query failed, and `runOrNull`'s `null` was read as `no-credentials`.

**A check that reports "nothing to verify" is indistinguishable from one that passes.** That's what made this quiet.

## Verified against the running database, both directions

```
old form -> ERROR:  column "configencrypted" does not exist
            LINE 1: SELECT configEncrypted FROM connection LIMIT 1
new form -> SxbTRV68DPLuPLxz:W/zCWqiEHhlG6y5m... (ciphertext returned)
```

## Fix

A new `runFileOrNull` — the argv counterpart to `runOrNull` — so nothing touches a shell. The docker branch was always fine: `dockerExec` passes its command to `docker exec` without the same interpolation.

## Why the original tests missed it

This is the part worth keeping. They mocked the exec helper and asserted only the **outcome mapping** — `ok` / `mismatch` / `no-credentials` / `unavailable`. Every one was green while the command itself was malformed. The mock made the module look tested; nothing ever executed the string.

The new test asserts the **argv**: `psql` as the file, the SQL in **one** entry, and `"configEncrypted"` still quoted inside it.

A related note on process: `tsc` caught my first attempt at this fix while the *tests still passed*, because they mock the module I was changing. Same shape, one layer up.

Found by an agent auditing #1278's neighbours, not by the suite.

356 CLI tests, typecheck, lint clean.

Refs #1274

### DIFF
--- a/cli/src/__tests__/lib/credential-probe.test.ts
+++ b/cli/src/__tests__/lib/credential-probe.test.ts
@@ -3,6 +3,7 @@ import { createCipheriv, randomBytes } from "node:crypto";
 
 vi.mock("../../lib/exec.js", () => ({
   runOrNull: vi.fn(),
+  runFileOrNull: vi.fn(),
   dockerExec: vi.fn(),
 }));
 
@@ -15,7 +16,7 @@ vi.mock("../../lib/config.js", () => ({
   getMode: vi.fn(() => "docker"),
 }));
 
-import { runOrNull, dockerExec } from "../../lib/exec.js";
+import { runOrNull, runFileOrNull, dockerExec } from "../../lib/exec.js";
 import { getMode, readProjectConfig } from "../../lib/config.js";
 import { probeCredentialDecryption } from "../../lib/credential-probe.js";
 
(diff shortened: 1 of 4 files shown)


alfredo1996:
Sonar failed at **0% on new code**. The three lines added to `exec.ts` had no coverage, because every test that reaches them mocks the whole `exec.js` module.

**That is the third instance today of a mock making code look tested** — and the least defensible one, because the credential-probe tests mocked *this same module* and asserted only the outcome mapping, which is precisely why the shell-quoting bug this PR fixes was invisible. Adding an untested helper while fixing that is not a great look.

Four cases now run against the real helper: trimmed stdout, null-not-throw on failure, options forwarded, and — the one that matters — **argv passed through unchanged**, with a quoted SQL identifier arriving as one argument with its quotes intact. Proven by mutating the helper to drop all but the first argument and watching the suite go red.

360 CLI tests, lint clean.

Issue #1284, opened 2026-07-27
fix(component): multi-selects mark the highlighted row aria-selected and filter search on option.value, not label


alfredo1996:
Two **independent root causes** (so: two fixes, not one) in the same family of cmdk-backed pickers. They land in the same files, which is why they are filed together.

## Problem

**Defect 1 — checked state is conveyed only by a decorative `<div>`, and cmdk's `aria-selected` actively reports the wrong item.**

`component/src/components/composed/multi-select.tsx:136-147`:

```tsx
<div
  className={cn(
    "mr-2 flex h-4 w-4 shrink-0 items-center justify-center rounded-sm border border-primary",
    value.includes(option.value)
      ? "bg-primary text-primary-foreground"
      : "opacity-50",
  )}
>
  {value.includes(option.value) && <Check className="h-3 w-3" />}
</div>
```

No `aria-checked`, no `aria-selected`, no `aria-multiselectable` — colour plus a checkmark glyph are the only carriers of state (WCAG 1.4.1, 4.1.2). Worse, cmdk already puts `aria-selected` on every `CommandItem`, bound to **highlight**, not selection. So the attribute is present and wrong: `true` for the merely-hovered row, `false` for every row the user actually checked.

**Defect 2 — `CommandItem` is given the machine value, so type-to-filter matches nothing.**

`component/src/components/composed/multi-select.tsx:130-134`:

```tsx
<CommandItem
  key={option.value}
  value={option.value}      // ← cmdk's built-in filter scores this string
  disabled={option.disabled}
  onSelect={() => handleToggle(option.value)}
>
```

The real consumer feeds UUIDs — `app/src/components/widget-editor/advanced-form-refresh-section.tsx:20-23`:

```ts
const options: MultiSelectOption[] = otherWidgets.map((w) => ({
  value: w.id,
  label: w.title || "(untitled)",
}));
```

with `searchPlaceholder="Search widgets…"`. In the widget editor's "After Submit → refresh these widgets" picker, typing any widget title collapses the list to the empty state; the only way to find a widget is to scroll.

The sibling `component/src/components/composed/combobox.tsx:106-108` is the **reference implementation** — it already fixes exactly this and documents why:

```tsx
// Use label as the cmdk filter value so typing the connection
// name (not the UUID) finds the right item.
value={option.label}
```

### Locations

| File:line | Filter value | Selection semantics |
| --- | --- | --- |
[...]

PR #1363, opened 2026-07-29
refactor(component): collapse CascadingSelector into ParamSelector — cascading selects are searchable now (#1360)

### LINKED ISSUE #1360: fix(component): cascading select has no search at all, and should arguably not be a separate widget type
## The bug

Type-to-search does not work in a cascading select. Not because it is broken — because **it was never built**.

`parameter-widgets/param-selector.tsx` (the plain select) is built on `Command`:

```
Command, CommandEmpty, CommandGroup, CommandInput, CommandItem, CommandList
searchable?: boolean   // "Enable search-as-you-type (shows a search input that triggers onSearch)"
onSearch?: (term: string) => void
```

`parameter-widgets/cascading-selector.tsx` is built on raw Radix `Select`:

```
Select, SelectContent, SelectItem, SelectTrigger, SelectValue
```

No `Command`, no `CommandInput`, no `searchable`, no `onSearch`. There is no code path by which typing could filter anything.

That makes it a **P2 that behaves like a P1 on real data**: a cascading select exists precisely to narrow a large list by a parent value, so it is the widget most likely to be pointed at hundreds of options — and it is the one you cannot type into.

## The design question, which is the better half of this issue

> why can't we just have the select that is cascading by default?

Comparing the two prop interfaces, they are the same component:

| `ParamSelectorProps` | `CascadingSelectorProps` |
|---|---|
| parameterName, options, value, onChange, placeholder, loading, className | parameterName, options, value, onChange, placeholder, loading, className |
| `searchable`, `onSearch` | — |
| — | `parentValue`, `parentParameterName` |

**Cascading is `param-selector` minus search, plus two optional parent props.** Nothing about "this list is filtered by another parameter" requires a different component — it requires two more optional props. The split is why search exists in one and not the other, and it will keep causing that: every future select feature has to be built twice or silently diverge.

It also leaks into the product. `cascading` is a distinct `paramUIType` in the widget editor (`widget-editor-store.ts:29`, `parameter-config-section.tsx:40`), so a user must decide **up front** whether a parameter is cascading, and changing their mind means switching widget type. Whether a select happens to be filtered by a parent is a property of its configuration, not a different kind of thing.

## Proposed direction

1. **Fix the bug now**: give `cascading-selector` the same `Command`-based body as `param-selector`, so search works. Smallest change that unblocks users.
2. **Then collapse the two**: add optional `parentValue` / `parentParameterName` to `param-selector`, make `cascading-selector` a thin wrapper (or delete it), and reduce `paramUIType` so "cascading" becomes a configuration of `select` rather than a sibling of it.

Doing (1) first is deliberate — (2) touches the widget editor, the store, the preview and stored dashboard configs, and should not block a fix for people who cannot use the widget today.

**Do not skip to (2) alone.** A migration that also has to introduce search is two risks in one change, and stored dashboards reference `cascading-select` as a widget type, so (2) needs a config migration path with the "no customers yet, schema can be reset" caveat weighed explicitly.

## Required tests

**Component (`component/src/components/composed/__tests__/cascading-selector.test.tsx` — exists already)**
- typing filters the visible options; a non-matching term shows the empty state
- `onSearch` fires with the typed term when `searchable` is set, and is NOT called when it is not
- **the parent-gating still works with search present**: with no `parentValue`, the control is disabled and shows `Select <parent> first…`; typing into a disabled control changes nothing
- selecting a filtered option calls `onChange` with the option's value, not its label — the `#1284` bug in `multi-select`, which handed cmdk the UUID instead of the label, is the same family and worth a regression case here

**For step 2, if taken**
- a `select` with `parentParameterName` set behaves identically to today's `cascading-select`
- a stored dashboard using the old `cascading-select` type still loads

**No E2E for step 1** — this is keyboard interaction in a popover, which jsdom covers well and Playwright covers slowly. Reconsider for step 2, where a stored-config migration is involved.

## Acceptance criteria

- [ ] Typing in a cascading select filters options
- [ ] Parent gating, disabled state and placeholder are unchanged by the fix
- [ ] The two components share one implementation rather than two that drift
- [ ] A decision is recorded on whether `cascading` remains a distinct `paramUIType`

alfredo1996:
## What

A cascading select could not be typed into, because `cascading-selector.tsx` used raw Radix `Select` while `param-selector.tsx` used `Command`+`CommandInput`. The data layer already supported both — `useSeedQueryOptions` takes `parentParameterName` **and** `searchable` (default `true`). The two components are now one.

`ParamSelector` gained optional `parentValue` / `parentParameterName`. When a parent is named it disables, labels "(depends on X)", and prompts "Select X first…" — in **both** the searchable and plain branches. `cascading-selector.tsx` and `param-cascading-select.tsx` are deleted.

Cascading is now keyed on *having a parent*, not on a widget type. `ParamUIType` loses `"cascading"`; the Parent Parameter field moves into the Select editor as optional. No back-compat shim for the old `cascading-select` widget type — pre-launch, per the project's standing position. A stored dashboard using it would fail to resolve its widget type.

## Two real bugs found by doing the merge

**1. A safeguard I claimed existed, didn't.** I stated in the issue that `use-seed-query-options.ts` already resets the search term on parent change. It does not — the reset keys on `searchable !== prevSearchable`. So the stale-filter bug was **live**: change the parent, and the previous search term stayed applied to a freshly-loaded option set. Fixed in the hook and in `form-widget-renderer.tsx`, which carries its own copy of the same logic.

**2. The merge introduced its own.** With the Parent field now on every select, clearing it writes `""` — and a `parentParameterName !== undefined` check reads that as a real parent, permanently disabling the select. Gating is on truthiness.

Also moved `aria-labelledby` from the Radix `<Select>` root (where it renders nothing) onto `<SelectTrigger>`, giving plain selects an accessible name.

## Proved by breaking — 12 mutants, all red

searchable-branch gating · radix-branch gating · parent prompt → generic placeholder · popover not closing on parent clear · `onChange` sending label instead of value (the #1284 family) · search input rendered when `searchable=false` · dependency hint dropped · empty parent name treated as real · search term surviving parent change · seed query not gated on parent · parent params not sent · child not cleared when parent moves.

**A process note worth keeping.** The first two mutation runs reported mutants as *passing*. The harness used package-relative paths against the worktree root, so `cmp` compared two nonexistent files, always reported "mutated", and vitest ran on unmutated source. It was caught only because a by-hand check contradicted it, and the harness was rebuilt with a canary that greps the source *during* the run. A mutation harness that cannot fail is worth less than no harness, because it manufactures confidence.

## Verification

Run in the **main tree with complete `node_modules`**, which matters: in the agent's worktree `tsc` was silently checking `app/` against the main repo's *old* `ParamSelector`.

- component **2068** tests / 203 files
- app **3309** tests / 253 files
- **E2E 332 passed, 0 failed** (15.6m)
- lint 0 errors, `tsc --noEmit` clean

Visual check in Storybook, both themes, both states: parent unset → dimmed trigger, "Select country first…", "CITY (depends on country)"; parent set → enabled, amber focus ring, **and a search input in the popover** — the thing that did not exist before.

## Found, not fixed

- **cmdk filters by `value`, not label.** A seed query returning `{value: uuid, label: "Italy"}` hides everything when you type "Italy". Shared with `ParamMultiSelector`; one-line remedy (`keywords={[opt.label]}`). Filing separately — it is the plain select's bug too, not this refactor's.
- Popover renders narrower than its trigger (`w-full` inside a portal). Pre-existing.
- `form-widget-renderer.tsx` duplicates `use-seed-query-options.ts` wholesale — debounce, parent params, gating, clear-on-parent-change. Kept in sync here; should be one hook.
- **Deleted three fake tests.** `param-cascading-select.test.ts`'s "parent params construction" block re-implemented the logic inside the test and asserted against its own copy. It touched no production code.
- Dark-mode `bg-accent` on the selected Command row is a solid amber-brown, much heavier than light. Pre-existing, affects every Command popover.

Closes #1360

## Summary by CodeRabbit

* **New Features**
  * Select parameters can now be configured as dependent on a parent parameter via **“Depends On”**.
  * Dependent controls prompt **“Select <parent> first…”** and remain disabled until the parent value is available.
  * Search and child selections reset automatically when the parent changes.
  * Added cascading examples for both searchable and non-searchable selectors in component preview stories.

* **Bug Fixes**
  * Legacy dependent configurations are handled as standard select behavior (no dedicated cascading-select mode).
  * Improved empty/loading messaging, filtering, and option list behavior for dependent selectors.

* **Documentation**
  * Updated demo/chart-gallery text to reflect the “Depends On” cascading model.


alfredo1996:
**Correcting the PR body: the E2E claim in it is wrong.**

I wrote "E2E 332 passed, 0 failed". The run reported:

```
19 failed
  [chromium] › e2e/parameter-types.spec.ts:467:7 › cascading-select refreshes child options…
5 skipped
332 passed (15.6m)
```

I grepped the output tail and took the last matching line. The line above it said `19 failed`. CI then found the same thing on shards 3 and 4 — CI was right, my report was not.

**The failures are real and belong to this change.** 19 specs across 7 files drive the `cascading-select` widget type that this PR removes:

| Spec | Failures |
|---|---|
| `transforms.spec.ts` | 5 |
| `parameters.spec.ts` | 4 |
| `form-widget.spec.ts` | 4 |
| `widgets.spec.ts` | 2 |
| `widget-library.spec.ts` | 2 |
| `parameter-types.spec.ts` | 1 |
| `dashboards.spec.ts` | 1 |

The author could not have caught this — the brief forbade running E2E, so the only signal that would have exposed the gap was unavailable to it. That is a defect in how I scoped the work, not in the work.

**This PR is not mergeable until those specs are migrated** to drive a select with a parent parameter instead of a `cascading-select` widget type. That migration is part of the change, not a follow-up: removing a user-facing widget type while leaving the suite driving it is what CI is for.

Holding the PR open and fixing it.

alfredo1996:
**Fixed — and both root causes were real defects, not test brittleness.**

**1. Two labels where one contains the other.** Moving the parent field into every Select editor put "Parameter Name" and "Parent Parameter Name" side by side. Ambiguous to any accessible-name lookup, and to a human scanning the form. Renamed to **"Depends On"**, which is clearer regardless of the test.

**2. The dependency hint was part of the accessible name.** It lived inside the child's `<Label>`, so the child's name computed as `movie(depends on year)` — embedding *another control's* name, making the two indistinguishable, and causing a screen reader to announce the dependency as part of the control's identity. Moved to `aria-describedby`: the name identifies, the description explains.

Both were reachable only through E2E. **The unit tests passed throughout**, because they assert on rendered text rather than on the accessible name the browser computes — worth noting, since that is the gap that let this reach CI.

**An over-correction of mine, reverted in the same commit.** I had also made six E2E locators `exact: true`. Once the rename removed the ambiguity that became unnecessary and actively wrong — the real accessible name is `Parameter Name *` with the required asterisk, so exact matching found nothing. That is what took the run from 19 failures to 6 rather than to 0: two fixes for one problem, the second undoing the first.

**Result:** 44 passed / 0 failed across `parameters`, `form-widget` and `dashboards` — the three specs that were failing. Checked the entire output for failure indicators, not just the summary line. Plus 2068 component, 3309 app, lint and typecheck clean.

**One gap I am flagging rather than fixing here:** `form-fields-editor.tsx:248` still gates the parent field on `parameterType === "cascading-select"`, which this PR retires from the editor. So a form field can have its parent *edited* if one exists, but not *set* on a new field. The stored type is still honoured at runtime (`parameter-store.ts`, `form-widget-renderer.tsx`), so nothing breaks — but the editor is inconsistent. Worth its own issue rather than growing this diff further.

alfredo1996:
**CodeRabbit review — 1 fixed, 2 declined with reasons.**

### Fixed: `parentParameterName` never reaches `ParamMultiSelect` (74f7d166)

This one was real, and this PR caused it. `Depends On` sits inside the same
editor block as the `Allow multiple selections` checkbox, so after #1360 a
cascading **multi**-select is a shape a user can save. `useSeedQueryOptions`
already withholds the seed query until the parent has a value for both
`select` and `multi-select` — but `ParamMultiSelect` never received the parent
name, so `ParamMultiSelector` had no way to say why it was empty. It rendered
an enabled dropdown that never filled, with no prompt and no hint.

Wired the parent through the three hops the select path already uses, and gave
`ParamMultiSelector` the same disabled state, `Select <parent> first…`
placeholder, and `aria-describedby` hint (a description, not part of the
accessible name — same split as dd61d48f).

The prop is optional at every hop, so dropping it is a silent regression the
type-checker cannot catch. Pinned at both layers: component behaviour in
`parameter-widgets.test.tsx`, and the wiring in a new renderer test that fails
if either hop is removed (verified — it fails on the pre-fix tree).

### Declined: `exact: true` on the `year` combobox locator

Stale as of dd61d48f, which landed after the review. The dependency hint moved
out of the accessible name into `aria-describedby`, so the child's name is
`movie`, not `movie (depends on year)`. There is no longer a name that
contains `year` other than `year` itself, so the default substring match is
unambiguous. Both parameter shards pass.

### Declined: compatibility path for stored `cascading-select` configs

The premise is right, the consequence isn't. A stored `cascading-select` does
fail validation and fall back to schema defaults, so it renders the
"No parameter name" empty state — that part is deliberate and documented in
`parameter-select/settings.ts`.

But it does **not** "re-save as default select settings and lose the parent".
`reverseParamTypeMapping` reopens the widget as the select it always was, and
`parentParameterName` lives in `chartOptions`, which the editor never
rewrites. Asserted in `widget-editor-store.test.ts:425` and
`parameter-config-section.test.tsx:150`.

No seeded dashboard is affected — `scripts/demo/*.json` and
`docker/postgres/seed-neoboard.sql` contain no widget of that type. The one
hit was a prose list in `chart-gallery.json` advertising `cascading-select` as
a supported type; that's corrected in the same commit.

Leaving the fallback as-is: pre-launch, no stored data in the repo depends on
it, and the empty state prompts exactly the re-save that fixes it. Worth
revisiting if we ever need to render an untouched pre-#1360 dashboard.

alfredo1996:
**Unit & Integration Tests failed on the last push. Root cause found — it is not this PR's feature change, and it is worth reading.**

CI reported `Test Files 254 passed (254)`, `Tests 3315 passed (3315)`, **exit 1**.
Nothing failed an assertion; vitest caught one unhandled error:

```
ReferenceError: window is not defined
 ❯ performWorkOnRootViaSchedulerTask react-dom-client.development.js
 ❯ Immediate.performWorkUntilDeadline scheduler.development.js
This error originated in "src/components/parameters/__tests__/use-seed-query-options.test.ts"
```

That attribution is a red herring — vitest blames whichever file was running
when the stray callback surfaced, not the one that queued it.

**What was actually wrong.** `vitest.config.ts` splits app tests by extension:
`*.test.ts` → `unit` (environment "node", **no setupFiles**), `*.test.tsx` →
`component` (jsdom + `vitest.setup.tsx`). `afterEach(() => cleanup())` lives
only in that setup file, so it only ever ran for `.test.tsx`.

Seven hook tests were `.test.ts` files that opted into jsdom with a
`// @vitest-environment jsdom` docblock. They got a DOM but no cleanup:
`renderHook` mounted React and nothing unmounted it. React's scheduler kept a
callback queued, vitest tore down the jsdom environment at end of file, and if
the callback fired after that teardown it threw. Whether it lands is a race —
which is why it failed on CI and passed locally, twice, including under
`test:coverage`. My new jsdom test shifted process timing enough to lose the
race; the leak has been there all along.

**Fix**: renamed those seven to `.test.tsx`, so they run in the project whose
environment they were already requesting and inherit cleanup. No test bodies
changed, and the now-redundant docblocks are gone. CLAUDE.md already documents
`unit` as "No DOM" — this restores the stated convention rather than inventing
one.

Also added a ratchet (`app/src/__tests__/test-environment-boundary.test.ts`)
that fails **by name** on any future `*.test.ts` that opts into jsdom or
imports `@testing-library/react`. Verified it fires: renaming one file back
produces `+ "hooks/__tests__/use-countdown.test.ts"`. Effectively all of the
diagnosis cost here was the misattribution, and the next one should be legible
on sight.

Local: `npm run verify` green across all four packages — app 3317, component
1690, connection 392, cli 69.

PR #1364, opened 2026-07-29
fix(graph): plain click single-selects a node instead of accumulating (#1191)

### LINKED ISSUE #1191: Graph: plain click accumulates node selection — should single-select (only Cmd/Ctrl+click should multi-select)
## Summary
In the interactive graph widget, a **plain click** on a node *adds* it to the current selection instead of replacing it. Every click accumulates, so you end up with several nodes selected. Because the property inspector only opens for a single selection, you can no longer inspect individual nodes once more than one is selected.

**Expected** (standard graph-tool convention):
- **Plain click** → select *only* that node (replace selection); clicking the already-sole-selected node deselects it.
- **Cmd+Click (macOS) / Ctrl+Click (Windows/Linux)** → toggle that node in/out of a multi-selection.

**Actual:** plain click and Cmd/Ctrl+click behave identically — both toggle the node into an accumulating set. There is no way to single-select, so the inspector becomes unreachable after the first extra click.

## Steps to reproduce
1. Open a dashboard with a graph widget (e.g. the demo **Chart Gallery → graph page**, or any Neo4j graph).
2. Click node A → its property inspector opens (selection = [A]).
3. Click node B (plain click, no modifier).
4. **Bug:** selection becomes [A, B]; the inspector panel closes (it only shows for a single node). You cannot inspect B on its own.

## Root cause
`component/src/charts/graph-chart.tsx` — the `onNodeClick` handler ignores the mouse event's modifier keys and unconditionally toggles the node into the selection set:

```ts
onNodeClick: (node) => {                       // (node, hitElements, event) — event dropped
  if (!onNodeSelectRef.current) return;
  const current = selectedRef.current ?? [];
  const id = node.id;
  const next = current.includes(id)
    ? current.filter((x) => x !== id)           // always additive/toggle
    : [...current, id];
  onNodeSelectRef.current(next);
},
```

NVL's callback signature is `onNodeClick(node, hitElements, event: MouseEvent)` (confirmed in `@neo4j-nvl/interaction-handlers` — the sibling `onNodeDoubleClick`/`onNodeRightClick` handlers already use the `event`). `event.metaKey` / `event.ctrlKey` are available but unused here.

Consequence surfaces in `app/src/components/graph-exploration-wrapper.tsx:267`, where the inspector opens only when `ids.length === 1` — so any second plain click closes it.

## Proposed fix
Branch on the modifier key:
```ts
onNodeClick: (node, _hit, event) => {
  if (!onNodeSelectRef.current) return;
  const current = selectedRef.current ?? [];
  const id = node.id;
  const additive = event.metaKey || event.ctrlKey;
  let next: string[];
  if (additive) {
    next = current.includes(id) ? current.filter((x) => x !== id) : [...current, id];
  } else {
    // plain click: select only this node, or clear if it was the sole selection
    next = current.length === 1 && current[0] === id ? [] : [id];
  }
  onNodeSelectRef.current(next);
},
```

## Required tests
- **Component (jsdom, `component/src/charts/__tests__/graph-chart.test.tsx`):** drive the `onNodeClick` callback with a mock `MouseEvent`:
  - plain click on B when [A] selected → `onNodeSelect(["B"])` (replace, not `["A","B"]`).
  - plain click on the sole-selected A → `onNodeSelect([])`.
  - `metaKey` click on B when [A] selected → `onNodeSelect(["A","B"])`.
  - `ctrlKey` click on an already-selected node → removed from the set.
- **E2E (`app/e2e/`, graph exploration flow):** on a seeded graph, click node A → inspector shows A; plain-click node B → inspector shows **B** (not closed); Cmd+click a third → two selected, inspector reflects multi-select state.

## Environment
Found while dogfooding the demo stack (release/1.3) ahead of a customer demo. Affects both macOS (Cmd) and Windows/Linux (Ctrl).

alfredo1996:
## What

`onNodeClick` in `component/src/charts/graph-chart.tsx` **dropped the event argument**, so modifier keys were invisible and every click unconditionally toggled the node into an accumulating selection. Since the property inspector only opens for a single selection, it became unreachable after the second click — you could not step through nodes one at a time, which is the most common thing anyone does with a graph.

Now: **plain click replaces** the selection (and clicking the sole selected node clears it); **Cmd/Ctrl/Shift toggles**.

## The signature was verified, not assumed

`node_modules/@neo4j-nvl/interaction-handlers/lib/interaction-handlers/click-interaction.d.ts:23`:

```ts
onNodeClick?: ((nodes: Node, hitElements: HitTargets, event: MouseEvent) => void) | boolean;
```

A real DOM `MouseEvent` — `metaKey`/`ctrlKey`/`shiftKey` all present. No window-level keydown tracking needed.

## Shift was checked, not assumed free

`InteractionHandlers.js` only constructs `BoxSelectInteraction`/`LassoInteraction` when `onBoxSelect`/`onBoxStarted`/`onLassoSelect`/`onLassoStarted` appear in `mouseEventCallbacks`. `graph-chart.tsx` passes none, and the repo has zero references to any of them, so neither handler is ever instantiated. NVL's only other modifier use is `ctrlKey||metaKey` for pinch-zoom, which is wheel-based. Shift therefore costs nothing, and accepting it alongside Cmd/Ctrl is friendlier than being right about the convention.

## A latent bug this also fixes

`graph-exploration-wrapper.tsx:264` and `plugins/graph/component.tsx:60` both fire click-action drill-downs with `ids[0]`. Under the accumulating selection, `ids[0]` stayed the **first-ever** node clicked — so after a second click, drill-downs were targeting the wrong node entirely. Corrected as a side effect. The blast radius of the reported symptom was wider than the inspector.

## Proved by breaking — six mutations, each reverted

| Mutation | Went red |
|---|---|
| drop `event?.metaKey` | both Cmd+click tests |
| drop `event?.ctrlKey` | both Ctrl+click tests |
| drop `event?.shiftKey` | both Shift+click tests |
| `next = [id]` (kill sole-selection deselect) | new test **and** the pre-existing "deselects if already selected" |
| `ids.length === 1` → `>= 1` in the wrapper | "closes the inspector when selection holds more than one" |
| inspector reads `nodes[0]` not `find(id === ids[0])` | "re-targets the inspector on the next single-id selection" |

Every assertion compares the resulting array (`toHaveBeenCalledWith(["2"])`), never bare `toHaveBeenCalled()` — which would have passed for every wrong behaviour here.

On mocks: the component test mocks only `@neo4j-nvl/react` (WebGL cannot run in jsdom); the code under test is untouched. The app test stubs `@neoboard/components` but exercises the wrapper's real `handleNodeSelect` — the last two mutations confirm it is not self-satisfying.

## Verification

component 2058 tests / 202 files · app 3309 tests / 255 files · lint clean · `tsc --noEmit` clean both packages. E2E ran green on this branch's base (332 passed).

**Visual check not performed** — nothing was listening on Storybook's port. Mitigating: the diff touches no rendering code (`selected: selectedIds.has(node.id)` is unchanged) and the three existing tests covering selection styling still pass. A `WithSelection` story exists in `graph-chart.stories.tsx` if you want to eyeball it.

## Not done

The issue also asks for an E2E leg covering the graph exploration flow. Not written here.

Closes #1191

### DIFF
--- a/app/src/components/__tests__/graph-exploration-wrapper.test.tsx
+++ b/app/src/components/__tests__/graph-exploration-wrapper.test.tsx
@@ -1,0 +1,130 @@
+/**
+ * GraphExplorationWrapper — property inspector wiring (#1191).
+ *
+ * The GraphChart itself is NVL/WebGL-backed and cannot render in jsdom, so it
+ * is stubbed here; the code under test is this wrapper's own `handleNodeSelect`
+ * — specifically the `ids.length === 1` rule that decides whether the property
+ * inspector opens. GraphChart's own click-to-selection semantics (plain click
+ * replaces, Cmd/Ctrl/Shift toggle) are covered where they live, in
+ * component/src/charts/__tests__/graph-chart.test.tsx. Together the two prove
+ * the chain that #1191 broke: a plain click yields exactly one id, and exactly
+ * one id opens the inspector.
+ */
+import { render, screen, act, cleanup } from "@testing-library/react";
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { GraphExplorationWrapper } from "../graph-exploration-wrapper";
+
+/** Props last handed to the stubbed GraphChart. */
+let chartProps: Record<string, unknown> = {};
+
+const graphNodes = [
+  { id: "A", label: "Alice", properties: { name: "Alice" } },
+  { id: "B", label: "Bob", properties: { name: "Bob" } },
+];
+
+const explorationStub = {
+  nodes: graphNodes,
+  edges: [],
+  selectedNodeIds: [],
+  onNodeSelect: () => {},
+  onExpandRequest: () => {},
+  collapse: () => {},
+  canExpand: () => false,
+  canCollapse: () => false,
+  reset: () => {},
+  expandedNodeIds: [],
+  expandingNodeId: null,
+};
+
+vi.mock("@neoboard/components", () => ({
+  GraphChart: (props: Record<string, unknown>) => {
+    chartProps = props;
+    return <div data-testid="graph-chart" />;
+  },
+  // Must be referentially stable across renders: the wrapper persists
+  // `exploration.nodes` / `.edges` to the store in an effect keyed on their
+  // identity, so a fresh object each render would loop forever.
+  useGraphExploration: () => explorationStub,
+  PropertyPanel: ({
+    sections,
+  }: {
+    sections: { title: string; items: { key: string; value: string }[] }[];
+  }) => (
+    <div data-testid="property-panel">
+      {sections.flatMap((s) =>
+        s.items.map((i) => (
+          <span key={`${s.title}-${i.key}`}>{`${i.key}=${i.value}`}</span>
+        )),
+      )}
+    </div>
+  ),
+  Badge: ({ children }: { children: React.ReactNode }) => (
+    <span>{children}</span>
+  ),
+}));
+
+function renderWrapper() {
+  render(
+    <GraphExplorationWrapper
+      widgetId="w1"
+      nodes={graphNodes}
+      edges={[]}
+      connectionId="c1"
+      settings={{}}
+      resultId="r1"
+    />,
+  );
+  return chartProps.onNodeSelect as (ids: string[]) => void;
+}
+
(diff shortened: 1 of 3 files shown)


alfredo1996:
@coderabbitai review

alfredo1996:
@coderabbitai full review

PR #1365, opened 2026-07-29
feat(graph): mark APOC virtual nodes and decline to expand them (#1361)

### LINKED ISSUE #1361: feat(graph): APOC virtual nodes render but cannot be expanded — negative-id nodes need an explicit unresolvable state
## What actually happens — measured, not assumed

Reported as *"I want to be able to resolve fake nodes using the negative ID. Currently this is not possible."*

I traced it against a live Neo4j 5 + APOC 5.26.25. **The transform is fine.** Given:

```cypher
MATCH (a) WITH a LIMIT 1
WITH a, apoc.create.vNode(['Summary'], {name:'Totals'}) AS v
RETURN a, v, apoc.create.vRelationship(a,'SUMMARISES',{},v) AS rel
```

over Bolt, an APOC virtual node carries a **stringified negative** `elementId`, and its virtual relationship points at it correctly:

| | `identity` | `elementId` |
|---|---|---|
| virtual node | `{low:-289, high:-1}` | `"-289"` |
| virtual rel | `{low:-275, high:-1}` | `"e1315589-…"` (a UUID) |
| virtual rel `startNodeElementId` | | `"-289"` ✓ |

Run through `transformToGraphData`'s id logic:

```
NODES:  4:1a7aa765-…:0[Document]    -290[Summary]
EDGES:  4:1a7aa765-…:0 -SUMMARISES-> -290
DANGLING EDGES: 0
```

`addNode` uses `elementId ?? identity`, `addEdge` uses `startNodeElementId ?? start` — they agree. **Virtual nodes render.**

## Where it breaks

`app/src/components/graph-exploration-wrapper.tsx:193`:

```ts
"MATCH (n)-[r]-(neighbor) WHERE elementId(n) = $nodeId RETURN n, r, neighbor"
```

A virtual node does not exist in the database, so this matches nothing. Verified against the live instance:

```
elementId(n) = "-290"        -> 0 rows
elementId(n) = <a real node> -> 876 rows
```

So expand **silently returns nothing**. The user clicks, waits, and gets no neighbours and no explanation — indistinguishable from "this node genuinely has no neighbours", which is the worst possible failure for an exploration feature. Nothing in the UI signals that this node is synthetic and can never be expanded.

## Proposed fix

**1. Detect and mark synthetic nodes at transform time.** A negative id is APOC's own convention precisely so virtual entities cannot collide with real ones — real Neo4j ids are never negative. Add `synthetic: true` to nodes whose resolved id parses as a negative integer, in `app/src/plugins/graph/transform.ts`.

**2. Make it visible.** Render synthetic nodes distinctly (dashed border is the usual convention) so a user can tell at a glance which parts of the graph are computed rather than stored.

**3. Decline expansion honestly.** Suppress the expand affordance on a synthetic node, or show *"This node was generated by the query (APOC virtual node) and has no neighbours in the database."* An explanation beats a silent empty result.

**Not proposed: making virtual nodes expandable.** They have no neighbours to fetch — the request is unsatisfiable, not unimplemented. If a user wants to explore *from* a virtual node they should expand the real node it was derived from.

## An honest caveat on the detection rule

`transform.ts:49` falls back to `crypto.randomUUID()` when a node has neither `elementId` nor `identity`, so "id parses as a negative integer" is a heuristic, not a guarantee. It is a good one — APOC chose negative ids for exactly this purpose — but it would misfire on a non-Neo4j connector whose ids happen to be negative integers.

Prefer detecting it where the information is certain if that is cheap: the parser sees the driver's own types. Worth 20 minutes of investigation before committing to the heuristic.

## Required tests

**Unit — `app/src/plugins/__tests__/` (graph transform)**
- a node with a negative `elementId` is marked `synthetic: true`; one with a real `4:uuid:123` elementId is not
- `-0`, `"-290"`, and a numeric `-290` all detect identically — the id arrives stringified
- a node whose id came from the `randomUUID()` fallback is NOT marked synthetic
- edges between a real and a virtual node still resolve, with **zero dangling** — this is the behaviour that already works and must not regress; fixture built from the real Bolt shapes in the table above, not invented

**Component — graph chart**
- a synthetic node renders with the distinct style; a real one does not
- the expand affordance is absent (or disabled with the explanation) on a synthetic node

**No E2E** — this needs APOC in the test Neo4j, and the assertion is on transform output plus a rendered attribute, both of which jsdom covers.

## Acceptance criteria

- [ ] A user can tell at a glance which nodes are synthetic
- [ ] Expanding a synthetic node explains why it cannot be expanded, rather than returning silently empty
- [ ] Mixed real/virtual graphs still render with no dangling edges
- [ ] The detection rule is documented, including what it would misfire on

alfredo1996:
## What was and was not broken

**The transform already handled APOC virtual nodes correctly** — ids line up, zero dangling edges. That behaviour is untouched and now pinned by a regression test.

**What broke was expansion.** `graph-exploration-wrapper.tsx:193` runs `MATCH (n)-[r]-(neighbor) WHERE elementId(n) = $nodeId`, which returns 0 rows for a virtual node — *silently*. Indistinguishable from "this node has no neighbours", which is the worst possible failure for an exploration feature.

Synthetic nodes are now marked at transform time, rendered distinctly, and the Expand affordance is replaced with an explanation. They are **not** made expandable — there is nothing to fetch; the request is unsatisfiable, not unimplemented.

## No certain signal exists — measured, not assumed

Probed the live Neo4j 5 + APOC 5.26.25 read-only and dumped the raw driver objects. A virtual node and a stored node are **the same class with identical shape**:

```
virtual: Node  ownKeys=[identity,labels,properties,elementId]  proto=[constructor,as,toString,__isNode__]
         elementId "-296"           identity {low:-296, high:-1}
real:    Node  ownKeys=[identity,labels,properties,elementId]  proto=[constructor,as,toString,__isNode__]
         elementId "4:1a7aa765-…:0"  identity {low:0, high:0}
```

The sign of the id is the only difference at **any** layer, so detecting it in `Neo4jRecordParser` would be no more certain — and strictly worse, because `parseGraphObject` returns `Path` and `PathSegment` **untouched**, so path-borne virtual nodes would go unflagged. Detection went in `addNode`, the single funnel every node/relationship/path shape routes through.

The rule, its misfire case (a non-Neo4j connector emitting `{labels, properties}` with genuinely negative ids), and the `-0` trap are documented in the code and in `docs/…/charts/graph.mdx`.

Edge cases, verified:

```
-297 → synthetic     -0 → not     0 → not     4:uuid:0 → not     -1 → synthetic
```

The `-0` exclusion matters: `String(-0) === "0"`, and node 0 is real.

## Proved by breaking — each restored after

| Break | Went red |
|---|---|
| remove the `synthetic` line | 4 detection tests |
| `/^-[1-9]\d*$/` → `/^-\d+$/` | the `-0` test |
| → `/-/` | real `4:uuid:0` **and** randomUUID-fallback tests |
| `startNodeElementId ?? start` → `start` | the zero-dangling-edges pin |
| `node.synthetic ?` → always | "does not mark a real node" |
| `if (node.synthetic)` → `if (!node.synthetic)` | "still expands real nodes" (+19 pre-existing) |
| wrapper `=== true` → `!== true` | 5 of 6 wrapper tests |

Fixtures use the **measured** Bolt shapes, including `identity` arriving as a number (the parser calls `Integer.toNumber()`) while `elementId` is a string — an invented fixture would have tested the assumption instead of the behaviour.

## Visual

NVL exposes no border/stroke option, so the dashed outline the issue proposed is unavailable. A synthetic node instead gets a second italic `virtual` caption line. Checked in both themes from a worktree-local Storybook (6006 serves the main repo and would not have shown the change): legible on the node fill, clearly distinct from a real node, and **not** error-coded. The context-menu explanation uses `text-muted-foreground`, asserted in a test never to be `destructive`.

## Environment caveats, stated plainly

`npm run test` and `tsc --noEmit` are red in the agent's worktree **regardless of this change** — verified by stashing (base 127 failed files, with-change 127 failed / 7 more passing). The component failures are the `storybook (chromium)` project resolving its setup file from the main repo's `node_modules`; the `tsc` errors are `@neoboard/components` symlinking to the main repo's `component/`, proven by temporarily repointing the tsconfig path. Both resolve in CI. Worth re-checking on this PR's own run rather than taking that on faith.

Closes #1361


alfredo1996:
@coderabbitai review

alfredo1996:
@coderabbitai full review

PR #1378, opened 2026-07-29
fix(component): a save on a narrow window permanently squashed the layout (#1375)

### LINKED ISSUE #1375: fix(dashboard): saving below a 1200px container permanently squashes the layout — widgets collapse toward one column
## Repro

1. Open a dashboard in edit mode on a normal laptop window.
2. Drag or resize **any** widget.
3. Save.
4. Widgets are narrower than before, and shifted toward a single column. Repeat
   and it gets worse each time. The original arrangement is **not recoverable** —
   it has been overwritten.

Reported as: *"sometimes when saving, the cards just go all in one column and they
become smaller."* The "sometimes" is the tell — it depends on window width.

## Root cause: one stored layout, but what gets stored is the current breakpoint's

`component/src/components/composed/dashboard-grid.tsx`:

```ts
const defaultBreakpoints = { lg: 1200, md: 996, sm: 768, xs: 480 };
const defaultCols       = { lg: 12,   md: 10,  sm: 6,   xs: 4  };

const layouts = { lg: layout, md: layout, sm: layout, xs: layout };   // same array
...
cols={{ ...defaultCols, lg: cols }}
onDragStop={handleUserLayoutChange}
onResizeStop={handleUserLayoutChange}
```

The **same** layout array is handed to every breakpoint, but each breakpoint has a
different column count. Below `lg`, react-grid-layout clamps every item into the
narrower grid: a 12-wide item becomes 10 at `md`, 6 at `sm`, 4 at `xs`.

`handleUserLayoutChange` then persists `currentLayout` — **the clamped layout for
whatever breakpoint is active** — through `updateGridLayout`
(`app/src/app/(dashboard)/[id]/edit/page.tsx:592` → `dashboard-store.ts:244`) and
into the single `gridLayout` stored on the page.

So the 12-column arrangement is **overwritten with its own squashed projection**.
Nothing ever widens it back, because the stored layout is the only one there is and
it is now literally narrower. Every subsequent save at a narrow width squashes it
again — a ratchet that only turns one way, ending at one column.

## Why it fires on ordinary windows

`width` is the **container**, not the viewport — the dashboard area minus the
sidebar (~240px) and padding. A 1280px window gives roughly 1000px of grid, which
is **below the `lg: 1200` breakpoint**, so normal editing already happens at `md`
(10 cols). Only a window wider than about 1500px stays on `lg`.

That is exactly the "normal editing is at the md breakpoint, don't gate on lg"
finding from the earlier grid-persistence work — the persistence path never
accounted for it.

## What the existing guard does and does not cover

The comment above `handleUserLayoutChange` already names this failure:

> Persist ONLY on real user drag/resize. `onLayoutChange` also fires on mount and
> on responsive reflow … forwarding those would dirty the dashboard and — since a
> single layout is stored — overwrite it with the reflowed narrow-column
> arrangement on save.

That correctly stops *incidental* reflow from persisting. It does **not** help
here: a real drag at `md` still hands back a 10-column layout, and that is
legitimately "a real user drag/resize". The guard fixed the trigger, not the
projection.

## Fix directions (for the drill)

- **(a) Author at one column count.** Give every breakpoint the same `cols` as the
  stored layout (12) and let the grid scale rather than reflow, so a drag can never
  produce a narrower layout. Smallest change; the trade-off is 12 columns on a
  narrow window.
- **(b) Store per-breakpoint layouts.** What RGL's responsive API actually expects.
  Correct, but it is a schema change to the stored dashboard and needs a migration
  plus a decision about which breakpoint a fresh dashboard authors in.
- **(c) Persist only when the active breakpoint matches the authoring one.** Keeps
  one stored layout and refuses to write a projection. Cheap, but silently drops
  the user's drag on a narrow window unless it is surfaced.

Whatever is chosen must make an existing squashed layout recoverable, or at least
stop making it worse — users already have dashboards that have ratcheted down.

Related inconsistency: `dashboard-store.ts:284` computes a new widget's position
with `x: (page.gridLayout.length * sourceGrid.w) % 12` — hardcoded to 12 columns
while the grid may be operating at 10, 6 or 4.

## Required tests

- **Unit (`app/`)** — `updateGridLayout` given a layout whose max `x + w` is
  narrower than the stored one does not silently replace it (whichever guard (a)–(c)
  is chosen, this is the invariant: **a save must never reduce the layout's column
  span**). Fails today.
- **jsdom (`component/`)** — `DashboardGrid` at a container width below 1200px:
  a simulated `onDragStop` does not emit a layout with items clamped to fewer
  columns than the input layout.
- **E2E (Playwright)** — at a 1280px viewport, open a dashboard, note each widget's
  `w`, drag one widget, save, reload, and assert every **other** widget's `w` is
  unchanged. This is the reported bug.
- **Regression** — repeat the save 3× and assert the layout is stable, not
  monotonically narrowing (the ratchet).

## Acceptance criteria

- [ ] Dragging/resizing and saving on a sub-1200px window leaves untouched widgets' geometry alone
- [ ] Repeated saves are idempotent — no progressive narrowing
- [ ] Widgets do not collapse to a single column
- [ ] New-widget placement no longer assumes 12 columns while the grid runs at another
- [ ] `npm run verify` green; E2E green

alfredo1996:
Closes #1375.

## The bug

One layout is stored per dashboard page, but `ResponsiveGridLayout` was handed
that single layout with **four different column counts** — lg:12, md:10, sm:6,
xs:4. Below `lg`, RGL clamps every item into the narrower count, and
`onDragStop` hands the clamped result back, which is then persisted as *the*
layout. The authored 12-column arrangement is overwritten by its own squashed
projection, and nothing ever widens it back, so each save on a narrow window
ratchets it further toward one column. Irreversibly.

The container is the viewport minus the sidebar, so a **1280px window already
measures below lg:1200**. It fired on ordinary laptops — hence "sometimes".

## The fix

One column count at every breakpoint. If you store one layout you have to author
at one column count; the grid now **scales** rather than reflowing, so a drag can
never return fewer columns than it was given.

Responsive stacking, if it is ever wanted, needs somewhere to store
per-breakpoint layouts. It cannot stay a side effect that destroys the only
layout we keep.

## What I did not do, and why

**No store-level "a save must never reduce the column span" guard**, which the
issue proposed as the invariant. At the store layer, a user legitimately resizing
widgets narrower is **indistinguishable** from a breakpoint projection — that
rule would reject real edits. The breakpoint context exists only in the grid, so
that is where the fix belongs.

**The existing `handleUserLayoutChange` guard is untouched** and was never
sufficient here. It correctly stops incidental reflow from persisting, but a real
drag at `md` *is* a real user drag and still returned a 10-column layout. It
fixed the trigger, not the projection.

**The `% 12` hardcode in `dashboard-store.ts:284` is left alone.** The acceptance
criterion was "new-widget placement no longer assumes 12 columns while the grid
runs at another" — that mismatch is now gone, because the grid always runs at the
authored count, and the app never passes a custom `cols`. Plumbing the value into
the store would be speculative.

## Tests

4 cases in `dashboard-grid.test.tsx`, verified **red before, green after**:

```
× uses the same column count at every breakpoint      AssertionError: expected 4 to be 1
× honours a custom cols at every breakpoint, not just lg
× keeps the authored column count at a narrow container width
→ 12 passed
```

`npm run verify` green across all four packages — app 3334, component 1712,
connection 392, cli 69.

## E2E: attempted, removed, and why

I wrote three separate E2E tests for this and **removed all of them**, because I
could not prove any of them catches the bug. Recording it rather than shipping a
green-looking test:

1. **API-based** (read the stored layout, compare widths) — failed on request
   plumbing, not on the bug.
2. **DOM-based at a fixed viewport** — passed *with and without* the fix. The
   stored layout and the breakpoint that renders it shrink together, so a
   `w:6-of-12` item saved as `w:5-of-10` renders at the same pixel width. Blind
   by construction.
3. **DOM-based wide→narrow→wide** — also passed without the fix. Tracing RGL's
   `correctBounds`: clamping to 10 cols keeps `w:6` and moves `x:6`→`x:4`, so
   items *overlap* and the vertical compactor stacks them. **Width is not what
   changes — position is.** And the seeded `Movie Analytics` fixture is already
   effectively single-column, so it cannot exhibit the symptom at all.

Doing this properly needs a purpose-built fixture: a dashboard with genuinely
side-by-side widgets, asserting **row/offset** rather than width. That is a
worthwhile follow-up and I did not want to gate a P1 data-loss fix on it.

The component test is the meaningful guard regardless — it pins the exact
invariant (`cols` identical at every breakpoint) at the only layer where
breakpoint context exists, and it fails the moment someone reintroduces a
per-breakpoint column map.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed responsive dashboard layouts so the configured column count remains consistent across all screen sizes.
  * Preserved custom column settings and authored 12-column layouts when viewed at narrower breakpoints.
  * Ensured dashboard layout positioning remains consistent across responsive breakpoints.

### DIFF
--- a/component/src/components/composed/__tests__/dashboard-grid.test.tsx
+++ b/component/src/components/composed/__tests__/dashboard-grid.test.tsx
@@ -125,3 +125,70 @@ describe("DashboardGrid — persist only on user drag/resize", () => {
     expect(capturedProps.resizeConfig.enabled).toBe(true);
   });
 });
+
+// ─── Breakpoint column parity (#1375) ────────────────────────────────────────
+//
+// A single layout is stored per page, but ResponsiveGridLayout was given four
+// DIFFERENT column counts (lg:12, md:10, sm:6, xs:4) for that one layout. Below
+// `lg` the grid clamps every item into the narrower column count, and
+// onDragStop hands that clamped layout back — which then gets persisted as THE
+// layout. The authored 12-column arrangement is overwritten by its own squashed
+// projection, and since nothing ever widens it back, every save on a narrow
+// window ratchets it further toward one column.
+//
+// The container is the viewport minus the sidebar, so a 1280px window is
+// already below the lg:1200 breakpoint — this fires on ordinary laptops, which
+// is why it read as intermittent.
+//
+// Fix: one column count everywhere. If you store one layout, you must author at
+// one column count; the grid scales instead of reflowing, so a drag can never
+// return fewer columns than it was given.
+
+describe("DashboardGrid — breakpoint column parity (#1375)", () => {
+  it("uses the same column count at every breakpoint", () => {
+    render(
+      <DashboardGrid layout={layout}>
+        <div key="a">A</div>
+      </DashboardGrid>,
+    );
+    const cols = capturedProps.cols as Record<string, number>;
+    const counts = Object.values(cols);
+    expect(new Set(counts).size).toBe(1);
+    expect(counts.every((c) => c === 12)).toBe(true);
+  });
+
+  it("honours a custom cols at every breakpoint, not just lg", () => {
+    render(
+      <DashboardGrid layout={layout} cols={24}>
+        <div key="a">A</div>
+      </DashboardGrid>,
+    );
+    const cols = capturedProps.cols as Record<string, number>;
+    expect(Object.values(cols).every((c) => c === 24)).toBe(true);
+  });
+
+  it("keeps the authored column count at a narrow container width", () => {
+    // 1000px is inside the md range (996–1199) — the width a 1280px window
+    // actually produces once the sidebar is subtracted.
+    mockWidth = 1000;
+    render(
+      <DashboardGrid layout={layout}>
+        <div key="a">A</div>
+      </DashboardGrid>,
+    );
+    const cols = capturedProps.cols as Record<string, number>;
+    expect(Object.values(cols).every((c) => c === 12)).toBe(true);
+  });
+
+  it("still hands the same layout to every breakpoint", () => {
+    render(
+      <DashboardGrid layout={layout}>
+        <div key="a">A</div>
+      </DashboardGrid>,
+    );
+    const layouts = capturedProps.layouts as Record<string, unknown>;
+    for (const bp of ["lg", "md", "sm", "xs"]) {
+      expect(layouts[bp]).toBe(layout);
+    }
+  });
+});
--- a/component/src/components/composed/__tests__/dashboard-grid.test.tsx
+++ b/component/src/components/composed/__tests__/dashboard-grid.test.tsx
@@ -151,10 +151,9 @@ describe("DashboardGrid — breakpoint column parity (#1375)", () => {
         <div key="a">A</div>
       </DashboardGrid>,
     );
-    const cols = capturedProps.cols as Record<string, number>;
-    const counts = Object.values(cols);
-    expect(new Set(counts).size).toBe(1);
-    expect(counts.every((c) => c === 12)).toBe(true);
+    // Exact mapping, not "all values equal": `{ lg: 12 }` alone would satisfy a
+    // set-size check while leaving md/sm/xs unmapped, which is the very bug.
+    expect(capturedProps.cols).toEqual({ lg: 12, md: 12, sm: 12, xs: 12 });
   });
 
   it("honours a custom cols at every breakpoint, not just lg", () => {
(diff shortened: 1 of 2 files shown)


alfredo1996:
Applied — you were right, and it mattered.

`new Set(Object.values(cols)).size === 1` is satisfied by `{ lg: 12 }` on its own, so the test would have passed a "fix" that mapped only `lg` and left md/sm/xs unmapped. That is precisely the bug, so the guard had a hole exactly where it needed not to.

All three now use exact object assertions, and the narrow-container case pins `capturedProps.width === 1000` as well, so it cannot silently stop exercising the sub-lg path if the mock changes. 12 passed.

Issue #1383, opened 2026-07-29
fix(e2e): dashboards.spec.ts duplicate-dashboard fails under parallel workers but passes in isolation


alfredo1996:
## Symptom

`app/e2e/dashboards.spec.ts:183` (duplicate-dashboard) fails on a full-suite run at
`--workers=2` and **passes in isolation**. Observed while verifying #1374: 352
passed, 1 failed. It fails **identically on the base tree** with that branch
stashed, so it is not caused by any change in flight.

Also seen at `--workers=4`: ~20 failures across the suite. Re-running the same 9
spec files gave 1 failure on base and 2 *different* failures on a branch, both
timing-sensitive `auto-refresh` tests. So the higher count is load-related, not a
code regression.

## Why this matters more than one flaky test

The project rule is **PERFECT-only merges to dev** — no flakes, no "known issues"
excuses. A test that fails only under parallel load quietly erodes that: every
future PR author has to decide whether their red run is theirs or this one, which is
exactly the judgement call the rule exists to remove.

It also interacts badly with local verification. This machine's Docker has **1 CPU**,
so parallel E2E already contends heavily (a run was lost this session to
testcontainer startup contention). A suite that is only green at low worker counts on
capable hardware is not a dependable gate.

## Likely cause

Cross-spec pollution rather than a race inside the test: it passes alone. Candidates
worth checking in order —

1. Shared dashboard **names** across specs, so a `getByText(name)` or a list-position
   assumption matches another spec's fixture. `dashboards.spec.ts` uses fixed names.
2. Dashboards created by other specs changing the list order or count that
   duplicate-dashboard asserts against.
3. Fixture cleanup that only runs on the happy path, leaving rows behind for whoever
   runs next in the same worker.

## Required tests

- Reproduce deterministically: run `dashboards.spec.ts` alongside the specific spec
  that pollutes it (bisect the file list) and record the pair.
- The fix should make the test **hermetic** — unique per-run dashboard names via
  `createTestDashboard` (`app/e2e/fixtures.ts:123`) and assertions scoped to its own
  fixture rather than to global list state — so it cannot depend on execution order.
- Verify green at `--workers=2` **and** `--workers=4` on a full-suite run, not just
  in isolation.

## Acceptance criteria

- [ ] `dashboards.spec.ts` passes in a full-suite run at 2 and 4 workers
- [ ] The test does not depend on dashboards created by other specs
- [ ] If other specs share the same weakness, they are named here for follow-up
[...]

PR #1462, opened 2026-08-03
fix(component): honour prefers-reduced-motion across all overlays (#1458)

### LINKED ISSUE #1458: fix(component): Dialog exit animation ignores prefers-reduced-motion, stalling under load — graph E2E flake and red dev
`dev` is red right now ([run 30818748876](https://github.com/alfredo1996/neoboard/actions/runs/30818748876)) and the same failure hits most dependabot PRs. It presents as "the graph chart E2E tests are flaky". It is not a graph bug — **the Add Widget dialog fails to unmount, and the graph tests are simply the ones heavy enough to expose it.**

## Evidence

Every failure across every run is the same assertion — `expect(dialog).not.toBeVisible()` — never a graph assertion:

| Run | Test | Line |
|---|---|---|
| dev (30818748876) | `graph chart — added widget renders on dashboard` | `charts.spec.ts:578` |
| dev (30818748876) | `graph chart — expand node loads neighbors` | via `addGraphWidget`, `charts.spec.ts:723` |
| PR #1457 | `graph chart — reset clears all expansions` | via `addGraphWidget` |
| PR #1457 | `graph chart — fullscreen dialog renders graph` | via `addGraphWidget` |
| PR #1453 | `graph chart — right-click shows context menu` | via `addGraphWidget` |

The Playwright call log is the proof:

```
Locator:  getByRole('dialog', { name: 'Add Widget' })
Expected: not visible
Received: visible
Timeout:  5000ms

4 × locator resolved to <div role="dialog" data-state="closed"
      class="... data-[state=closed]:animate-out data-[state=closed]:duration-150
             data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95 ...">
      - unexpected value "visible"
```

`data-state="closed"` — React state already flipped, Radix already began the exit. But the node is **still in the DOM and still visible after 5 000 ms**, for a 150 ms animation.

## Root cause

Radix keeps `DialogContent` mounted until the CSS exit animation fires `animationend`. In `addGraphWidget` the click that closes the dialog happens while:

- an NVL/WebGL graph preview is live **inside** the dialog and tearing down, and
- a second NVL graph widget is mounting on the dashboard grid behind it.

NVL saturates the main thread at exactly that moment (the run log also shows `[mobx] Warning: there are multiple mobx instances active` right at these tests). The 150 ms exit animation stalls, `animationend` never lands inside the budget, and Radix never unmounts.

That explains every symptom:

- **Only the graph block flakes** despite 44 other `expect(dialog).not.toBeVisible()` assertions in the suite — no other chart type mounts WebGL.
- **The failing test rotates** (`:553`, `:745`, `:772`, `:821`, `:946`) — it's whichever graph test lands while the runner is loaded.
- **Retries don't help** — the retry starts on a still-loaded runner.
- **It's load-dependent** — dev passes roughly two runs in three.

## The underlying defect is an accessibility bug

`component/src/components/ui/dialog.tsx` animates on both `DialogContent` (line 13) and `DialogOverlay` (line 51) with **no `motion-reduce:` guard** — unlike every other animated component in the library, which all guard correctly:

- `component/src/components/ui/spinner.tsx:12` → `motion-reduce:animate-none`
- `component/src/components/ui/progress.tsx:31` → `motion-reduce:animate-none`
- `component/src/components/ui/skeleton.tsx:19-21` → `motion-safe:` prefixes

So Dialog ignores `prefers-reduced-motion` today. That is a real a11y defect independent of CI, and fixing it fixes the flake in the same line.

## Proposed fix

1. Add `motion-reduce:animate-none` to the content and overlay variants in `component/src/components/ui/dialog.tsx` (brings Dialog in line with spinner/progress/skeleton).
2. Set `reducedMotion: "reduce"` in the `use:` block of `app/playwright.config.ts`.

With (1) in place, (2) makes the exit animation a no-op in CI, so Radix unmounts synchronously and the stall cannot occur regardless of main-thread pressure — for all 44 assertions, not just the graph ones. No per-test changes, no arbitrary timeout inflation.

Note that (2) alone does **not** work: without (1) the dialog animation ignores the reduced-motion preference entirely.

## Required tests

**Component (`component/`, Vitest + jsdom)**
- `dialog.test.tsx`: assert `DialogContent`'s class list contains `motion-reduce:animate-none`, mirroring the existing assertions in `component/src/components/ui/__tests__/spinner.test.tsx:34` and `progress.test.tsx:25`.
- Same assertion for `DialogOverlay`.

**E2E (`app/e2e/`, Playwright)**
- After the config change, run `charts.spec.ts` graph block at the default 2 workers and confirm a clean pass.
- Re-run shard 1 several times (`--repeat-each=3` on the `Graph chart` grep) to confirm the stall no longer reproduces.
- No new E2E test is warranted — the existing five graph tests already assert the behaviour; they just needed to stop racing an animation.

**Regression guard**
- Keep `expect(dialog).not.toBeVisible()` as-is. If it starts failing again, that is a genuine signal rather than the animation artefact it is today.

alfredo1996:
Closes #1458

## What this is

Filed as "the graph chart E2E tests are flaky". It is not a graph bug and it is not really a test bug — the ten Radix overlay primitives never honoured `prefers-reduced-motion`, and the CI flake was a symptom of that.

## Why the obvious fix doesn't work

`spinner`, `progress` and `skeleton` guard their animations with `motion-reduce:animate-none`. Applying the same to the overlays fails silently:

| Selector | Compiles to | Specificity |
|---|---|---|
| `data-[state=open]:animate-in` | `.data-\[state\=open\]\:animate-in[data-state=open]` | (0,2,0) |
| `motion-reduce:animate-none` | `.motion-reduce\:animate-none` inside `@media` | (0,1,0) |

A media query contributes no specificity, so the guard loses at every one of the fourteen animated surfaces. Only `!important` inside the media block wins — and doing that once beats doing it fourteen times and forgetting it on the fifteenth.

The reset lives in `component/design-tokens.css`, the only stylesheet that both `component/src/index.css` and `app/src/app/globals.css` import.

## Why `animation: none` and not `animation-duration: 0.01ms`

The common snippet shortens the animation. That keeps `animationName` non-`none`, and Radix's presence machinery then still waits for an `animationend` before unmounting. `animation: none` makes Radix unmount **synchronously**, with no event to miss.

That distinction is the CI fix. Radix keeps an overlay mounted until its exit animation reports back; when an NVL/WebGL widget mounts alongside the close, the 150 ms animation stalls and the dialog stays in the DOM well past any budget:

```
Locator:  getByRole('dialog', { name: 'Add Widget' })
Expected: not visible
Received: visible
Timeout:  5000ms

4 × locator resolved to <div role="dialog" data-state="closed" ...>
      - unexpected value "visible"
```

`data-state="closed"` — React already flipped, Radix already began the exit — yet still visible after 5 000 ms, for a 150 ms animation.

Every failure on `dev` and on the dependabot PRs was that same `expect(dialog).not.toBeVisible()` assertion, never a graph assertion. That is why the failing test rotated across the block (`:553`, `:745`, `:772`, `:821`, `:946`), why retries never helped, and why only the graph tests were affected despite 44 identical assertions elsewhere in the suite — nothing else mounts WebGL.

## Handled consequence

`scrollAndHighlight` removed its highlight class **only** on `animationend`. Under `animation: none` that event never fires and the class would have stuck forever. It now also listens for `animationcancel` — an interrupted pulse dispatches that instead, a latent leak that existed before this change — and keeps a timer as the backstop.

## Tests

The existing mock-based `scrollAndHighlight` tests asserted that a stubbed `addEventListener` had invoked its own callback, which stayed green whether or not the highlight was ever cleared. Replaced with jsdom tests against a real DOM (real `classList`, real events, real timers); the node file keeps the parts that genuinely need no DOM (`CSS.escape`, RAF retry budget).

`reduced-motion.test.ts` guards the CSS block against deletion or being defanged back to `0.01ms`. It is a text-presence guard and says so — the cascade itself is proven by the E2E suite, which now runs with `contextOptions: { reducedMotion: "reduce" }`, so the accessibility branch is exercised in CI rather than merely declared.

## Verification

- `npm run verify` — 0 errors (2 pre-existing `widget-editor-modal` warnings untouched)
- Full E2E — **359 passed, 0 failed.** 2 flaky that recovered on retry: `dashboards.spec.ts:183` (known #1383) and `grid.spec.ts:50`; neither involves a dialog
- Graph block `--repeat-each=3` — **30/30, zero flakes**

Before this change the same command reproduced the stall reliably.

## Note for reviewers

No visual change for anyone not requesting reduced motion — the rules are inert outside the media query. For those who do request it, animations now stop, which is the point.

## Summary by CodeRabbit

* **Accessibility**
  * Added reduced-motion support that disables animations, transitions, and smooth scrolling when the setting is enabled.

* **Bug Fixes**
  * Improved widget highlighting cleanup, including animation cancellation and fallback handling when animations are disabled or do not complete.

* **Tests**
  * Added coverage for reduced-motion behavior, widget highlighting, cleanup, hidden widgets, and stale timers.


alfredo1996:
Both review findings addressed in `482a0a43`, plus a third issue they led me to.

**Bubbling animation events — fixed, good catch.** Real defect: a widget card is full of independently animating content (skeletons, spinners, ECharts), and the first descendant animation to finish would have stripped the highlight. The handler now runs only when `event.target === el`, with a jsdom test dispatching bubbling events from a child.

**Test location — fixed.** Moved to `app/src/lib/widget/__tests__/scroll-to-widget.test.ts`, node environment preserved. While moving it I also stopped its fake element invoking listener callbacks with no argument — no real listener does that, and it was the only thing masking the bubbling bug.

**On the "Linked Issues" pre-merge warning** (missing `DialogContent`/`DialogOverlay` reduced-motion class assertions): that check is comparing against the test plan I wrote in #1458, which proposed per-component `motion-reduce:animate-none`. I abandoned that approach during implementation because it cannot work — those classes would lose the cascade, as set out in the PR description. There are no per-component classes to assert. The equivalent guard is `component/src/__tests__/reduced-motion.test.ts`, which pins the global block including the `animation: none` vs `0.01ms` distinction that actually matters. I have updated #1458 so the issue no longer describes a superseded plan.

**Separately, `dc5bb6f3`.** Verifying this PR surfaced a genuine regression it caused: `charts.spec.ts:872` went from 5/5 green on `release/1.5` to 1 failure in 5 on this branch. Not a flake — attributable, so I chased it.

Cause: `await expect(Save).toBeEnabled()` is not a wait. The button is enabled before and after the mutation, so it passes on the first poll and the test races on while the store is still dirty; "Back" then trips the unsaved-changes guard whose overlay eats the pointer events. Animation timing had been hiding it. Notably `parameters.spec.ts` and `form-widget.spec.ts` had each already hit this and worked around it locally by clicking "Leave" if it appeared — one calls it a "grid compaction race". The dialog was never the problem.

`saveDashboard()` in fixtures now waits for the "Dashboard saved" toast, which `dashboard-workspace.tsx` raises only after `markSaved()`. Applied to the three Save-then-Back sites that had no workaround.

`charts.spec.ts:872` is now 6/6 with retries disabled, and the full suite is 360 passed / 0 failed.

alfredo1996:
Third finding addressed in the latest commit — correct, and the strongest of the three.

The assertions searched the whole stylesheet, so they would have passed on an `animation: none !important` living anywhere else in the file, including one applied unconditionally. That is the exact opposite of what the block is for, and the test would have said nothing.

They now run against the extracted media block only — 179 characters of 9,240 — plus one check that the extraction has not silently captured the whole file, which would otherwise make every assertion above it pass vacuously.

All three findings resolved. `npm run verify` green, full E2E 360 passed / 0 failed.

PR #1467, opened 2026-08-04
fix(charts): percent format scales ratios instead of appending a sign (#1396)

### LINKED ISSUE #1396: fix(charts): numberFormat 'percent' appends % without scaling — every percentage is 100x too small
## Symptom

A KPI computing `401 / 2000 = 0.2005` with `numberFormat: "percent"` renders
**`0.2%`**. The true share is **20.05%** — wrong by a factor of 100, rendered
cleanly, with no hint anything is off.

Observed on Chart Reference → Single Value.

## Cause

`component/src/charts/chart-utils.ts:68-73` — the `percent` branch appends a `%`
sign and does nothing else:

```ts
case "percent":
  formatted =
    decimalPlaces !== undefined
      ? `${value.toFixed(decimalPlaces)}%`
      : `${value}%`;
  break;
```

So `percent` silently means "this number is already 0–100". Nothing says so. The
option's description reads only *"…or percentage"*, and the seeded reference tile
is itself authored with **ratio** semantics (titled `(0.42)`), i.e. whoever wrote
the demo assumed the other convention — which is the clearest evidence the
contract is not discoverable.

## Why P1

It produces a **plausible wrong number** on the most prominent widget type in the
product. `0.2%` and `20.05%` are both believable share figures; nothing in the UI
distinguishes a formatting bug from a real value. Same class as v1.4's pie Top-N
(#1287) and the Neo4j temporal 10^6 error (#1306) — renders cleanly, reads wrong.

## Fix — decide the contract, then make it enforceable

Two defensible options; pick one and say so in the option description:

1. **`percent` means "ratio, scale it"** — multiply by 100. Matches
   `Intl.NumberFormat`'s own `style: "percent"`, and matches what the demo author
   assumed. Breaking for anyone who worked around it by pre-multiplying.
2. **`percent` means "already 0–100"** — keep the maths, but state it in the
   description and in the docs.

Recommend (1), using `Intl.NumberFormat(style: "percent")` so the behaviour is the
platform's rather than hand-rolled — and it picks up locale grouping for free.
Whichever is chosen, the description must state the expected input range; the
current wording is what allowed the demo to be authored wrong.

## Required tests

- **Unit (`component/`)** — `formatNumber(0.2005, "percent")` returns the agreed
  string; a boundary case (`1` → `100%` or `1%`) pins the convention so it cannot
  be flipped silently later.
- **Unit** — `decimalPlaces` interacts correctly with the scaling.
- Fix the seeded reference tile so the demo demonstrates the real contract.

## Acceptance criteria

- [ ] A ratio and a 0–100 value cannot both silently render as a plausible percentage
- [ ] The option description states the expected input range
- [ ] The seeded Chart Reference tile matches the documented contract
- [ ] `npm run verify` green

Found during the v1.5 chart review.

alfredo1996:
Closes #1396

> [!WARNING]
> **Breaking.** Queries that worked around the old behaviour by pre-multiplying by 100 now render 100x too large. Pre-launch, no customers — flagged rather than versioned.

## The bug

`numberFormat: "percent"` appended a `%` and scaled nothing. A KPI computing `401 / 2000 = 0.2005` rendered **`0.2%`** where the true share is **`20.05%`** — wrong by a factor of 100, rendered cleanly, with nothing in the UI to distinguish it from a real value. Both `0.2%` and `20.05%` are believable share figures.

## The root cause is the contract, not the arithmetic

The convention was never stated anywhere, and the two option surfaces documented **opposite** readings:

| Surface | Said |
|---|---|
| `chart-options/single-value.ts` | "…or percentage" — silent on range |
| `chart-options/table.ts` | `Percent (12%)` — asserts input is already 0–100 |
| Seeded reference tile | authored as a **ratio** (`COUNT FILTER / COUNT`, titled `(0.42)`) |

One `formatNumber` served all three. That the demo author picked the other convention is the clearest evidence the contract was not discoverable — so this PR changes the wording everywhere the option appears, not just the maths.

## The fix

`percent` takes a **ratio** and scales it, via `Intl.NumberFormat`'s own `style: "percent"` — the platform's behaviour rather than a hand-rolled one, which also brings locale grouping the old branch never had (`12.5` → `1,250%`).

**One judgment call the issue didn't cover:** with `decimalPlaces` unset, Intl's percent default is **0** fraction digits, which would render `0.2005` as `"20%"` — silently dropping precision, the same failure mode this fix exists to remove. Capped at `maximumFractionDigits: 2` instead, so `0.2005 → "20.05%"` and `0.5 → "50%"` with no padding. Commented at the call site.

The seeded tile already queried a genuine ratio, so it becomes **correct for free** — only its title needed to state the contract.

## Tests

8 new cases. Two pre-existing tests encoded the *wrong* contract and are inverted:

- `formatNumber(75, "percent") === "75%"` → `formatNumber(0.2005, "percent") === "20.05%"`
- `<SingleValueChart value={75} numberFormat="percent" />` → `value={0.75}`

The boundary case (`1` → `"100%"`, not `"1%"`) pins the convention so it cannot flip back silently.

## Verification

`component` 2120/2120 · typecheck + lint clean · chart E2E specs 48 passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Percentage formatting now correctly converts ratio values, such as `0.75`, to `75%`.
  * Supports decimal precision, zero-padding, thousands separators, negative values, and prefixes or suffixes.
  * Percentage values are capped at two decimal places when precision is not specified.

* **Documentation**
  * Updated chart option descriptions and examples to explain ratio-based percentage formatting.
  * Added guidance for providing ratio values in SQL and clarified compatibility considerations.

* **Tests**
  * Expanded coverage for percentage formatting and single-value charts.


alfredo1996:
@coderabbitai review

alfredo1996:
@coderabbitai full review

alfredo1996:
@coderabbitai full review

Issue #1425, opened 2026-07-31
feat(charts): introduce a field/series configuration tier — formatting defined once, applied everywhere


alfredo1996:
## Problem

Every option in the product is **panel-level** — it configures one widget as a
whole. There is no tier for "how this *data* should be formatted", and no tier
for "…except this one series".

That single missing layer is the root of several separately-filed bugs:

| Symptom | Filed as | Why the missing tier caused it |
|---|---|---|
| `decimalPlaces` advertised on bar, line, pie and dead on all three | #1397 | copy-declared into five per-chart arrays instead of defined once |
| A number reads differently in a chart label vs a KPI card | #1248 | no shared formatting definition to route through |
| `groupBy`/`calculatedColumn` emit `179.78999999…` | #1415 | no field-level precision to apply at render |
| `rightAxisSeries` is a **text input listing series names** | — | a per-series override with nowhere to live |
| The same styling rule means per-item on bar, series-mean on radar, series-last-value on line | #1417 | thresholds are re-implemented per chart instead of being one field-level concept |

Each was filed as its own defect. They are one design gap seen five times.

## Prior art

Grafana splits panel configuration three ways, and this is the split we lack:

1. **Panel options** — the visualization as a whole (legend, orientation, stacking)
2. **Standard field options** — unit, decimals, min/max, colour scheme,
   thresholds, value mappings — defined **once**, applied to every panel type
3. **Overrides** — per-field exceptions ("this series: right axis, red, 2 decimals")

Tier 2 is precisely where "the same number reads the same way everywhere" lives.

## Proposal

Introduce a field tier, in two stages so it can land incrementally.

**Stage 1 — hoist formatting.** Move `numberFormat`, `decimalPlaces` and a new
`unit` out of the per-chart option arrays into one **Data formatting** section
that every chart type gets automatically, and have the renderer apply it
uniformly through the existing `formatNumber` helper. This alone:

- deletes the duplicated declarations that produced #1397's `decimalPlaces` trio
- gives #1248 a single place to be fixed
- gives #1415 somewhere to apply precision at render rather than at transform

**Stage 2 — overrides.** A per-series list where a named series can override
selected field options. `rightAxisSeries` stops being a comma-separated text
input and becomes "axis: right" on one series; rule-based styling becomes a
field-level threshold set with an explicit, uniform unit of application, which
is what #1417 asks for.

## Scope and risk

This is the largest of the four proposals and the only one touching stored
[...]

PR #1468, opened 2026-08-04
fix(charts): reject long-format results in bar and line (#1400)

### LINKED ISSUE #1400: fix(charts): a long-format query renders plausible garbage in bar and line, with no warning
## Symptom

A long-format result — `category, series, value`, the natural shape of
`GROUP BY a, b` — renders a chart that looks plausible and is wrong.

**Bar** (Chart Reference, `stackMode = stacked` and `= percent`, querying
`parent.name, o.status AS series, SUM(...) AS revenue`): 12 bars with duplicated
x labels (Apparel, Apparel, Electronics, Home, Home, Outdoors), a ghost `series`
legend swatch with no bars, and **no stacking at all**.

**Line** is worse (`endLabel`, `right Y-axis series`, `showLegend = false`,
`palette = observable`, querying `week, status AS series, revenue`): it draws
**one** revenue line across duplicated x values, interleaving
delivered → shipped → delivered → shipped. The result is a violently spiky
"time series" that is pure artifact — and nothing marks it as such.

## Cause

`resolveValueKeys` treats **every** non-label column as a value series
(`app/src/plugins/transforms/shared-utils.ts:40-50`), so the `series` column
becomes a series in its own right. `toSeriesNumber` then turns `'delivered'` into
`null` (`:76-83`), which is why the legend entry exists with no bars.
`validateBarData` only counts columns (`app/src/plugins/bar/transform.ts:51-58`),
so nothing rejects it.

## Why P1

The chart renders cleanly. Duplicated category labels and a spiky line look like
*data*, not like a configuration error — and long format is the shape most SQL and
Cypher `GROUP BY` queries naturally produce. Same family as the pie Top-N bug
(#1287) and the temporal 10^6 error (#1306): renders fine, reads wrong.

That the seeded reference dashboard itself demonstrates two options with this exact
query shape is evidence of how easy it is to fall into.

## Fix — two defensible options

1. **Reject it loudly.** Fail validation when a chosen value column is
   non-numeric, and say which column and why.
2. **Support it.** Pivot long → wide on the series column, which is what the user
   plainly meant.

(2) is the better product outcome and is a real feature; (1) is the smaller change
and stops the wrong answer today. Recommend shipping (1) now regardless, since even
with (2) a genuinely non-numeric value column must still be rejected rather than
silently coerced to `null`.

## Required tests

- **Unit** — `validateBarData` / `validateLineData` reject a result whose selected
  value column is non-numeric, naming the column. Fails today.
- **Unit** — `toSeriesNumber` returning `null` for a whole column is surfaced, not
  swallowed.
- **jsdom** — a long-format result renders an explicit error state rather than a
  chart.
- Fix the seeded Chart Reference tiles so the stacking options demonstrate real
  stacking.

## Acceptance criteria

- [ ] A long-format result never renders as a silently wrong chart
- [ ] The message names the offending column
- [ ] Reference tiles for `stackMode` actually demonstrate stacking
- [ ] `npm run verify` green

Found during the v1.5 chart review.

alfredo1996:
Closes #1400

## The bug

A long-format result — `category, series, value`, what `GROUP BY a, b` naturally produces — rendered a chart that looked plausible and was wrong.

**Bar:** 12 bars with duplicated x labels, a ghost `series` legend swatch with no bars, and **no stacking at all** despite `stackMode` being set.
**Line:** worse — a single revenue line drawn across duplicated x values, interleaving delivered → shipped → delivered, which reads as a violently spiky time series but is pure artifact.

`resolveValueKeys` treats every non-label column as a value series, so `series` became a series of its own; `toSeriesNumber` turned `'delivered'` into `null`, which is why the legend entry existed with no bars. `validateBarData` only counted columns, so nothing rejected it.

## The fix — reject, don't pivot

`validateNumericValueColumns` in `transforms/shared-utils.ts`, wired into both validators. Rejection rule: a column is flagged only when it has **at least one non-null cell and none that parse**. An entirely null column stays legal — that is a sparse series, what a `LEFT JOIN` produces, and the reason `collectAllKeys` unions keys in the first place.

Resolution goes through `resolveValueKeys` — the same call the transform makes — so the validator cannot disagree with what is actually plotted.

Pivoting long→wide (the issue's option 2) is deliberately **not** here: it's a new feature against the no-new-features directive and overlaps #1425, already in v1.6. Even with a pivot, a genuinely non-numeric value column must still be rejected rather than coerced to null, so this lands either way.

## Two things the issue didn't anticipate

**1. `validate` couldn't see the column mapping.** Its signature was `(data: unknown)`, so a validator had to assume positional defaults and would wrongly reject a result whose text columns the user had already mapped away. `columnMapping` was already in scope at both `card-container` call sites — used on the very next line for the transform — so the signature is widened to `(data, mapping?)` to match `transformWithMapping`.

**2. 11 reference tiles used this exact shape** (5 bar, 6 line). Left alone they would have gone from *wrong chart* to *"Incompatible data format"*, which is worse for a reference dashboard. All 11 are rewritten to wide format with `SUM(...) FILTER (WHERE ...)`, so the `stackMode` tiles finally demonstrate real stacking. Side effect: `page-line-v-28` sets `rightAxisSeries: "shipped"`, a column that did not exist under the old query — that tile starts working too.

## Tests

19 new: 10 for the helper (including both mapping paths and the sparse-column cases), 3 per validator, 3 jsdom.

> [!NOTE]
> **jsdom coverage of any validator was fictional before this PR.** `chart-helpers` registers lightweight stub plugins when the real plugin modules haven't loaded, and `LightDef` carries no `validate` field — so `chartConfig.validate` is `undefined` in every jsdom test. My first attempt passed a long-format result straight through to a rendered chart for exactly that reason. The test now attaches the real `validateBarData` via `importActual` and says why; importing the plugin module itself would drag ECharts into jsdom for no benefit.

## Verification

`app` 3499/3499 · typecheck + lint clean · chart E2E (charts, chart-export, heavy-widgets, widget-editor) 48 passed, 3 skipped, 1 flaky — `chart-export` PNG-download timing, unrelated spec, green on retry.

## Follow-up worth filing

`ColumnMapping.groupBy` exists to designate a series column but `resolveValueKeys` only excludes the label key — so even a user who correctly declares `groupBy: "series"` gets it plotted as a series. This PR rejects that case with a clear message rather than rendering garbage, which is the right outcome while no pivot exists. Wiring `groupBy` up properly belongs with the long→wide pivot in #1425.

## Summary by CodeRabbit

* **Bug Fixes**
  * Bar and line charts now detect non-numeric plotted values and identify the affected columns.
  * Validation respects selected chart mappings, improving support for explicitly configured axes and value fields.
  * Entirely null sparse series remain valid.
  * Clear guidance is provided when long-format data requires conversion to wide format.

* **Documentation**
  * Added an Unreleased changelog entry describing the chart validation improvements.

* **Examples**
  * Updated reference chart tiles to use separate status series for clearer stacked, legend, palette, and dual-axis visualizations.


alfredo1996:
@coderabbitai review

alfredo1996:
@coderabbitai full review

alfredo1996:
@coderabbitai full review

alfredo1996:
@coderabbitai full review

